PR #24, opened 2019-02-05
Cleaner, more elegant logging

ubidefeo:
**ATTENTION**: requires consulting the Cloud team before merging
I have refactored the logging entirely and added one extra level of verbosity (4).
It is now leaner and cleaner.
I also implemented firmware version check and compare but have an issue with WiFiNINA reporting the incorrect version (1.2.0) as latest.


ubidefeo:
hey @lxrobotics 
thanks for taking a look.
That method is being rewritten entirely because its place is definitely not as an inline in that Class.
I guess this will have to be merged before I finish that, but for now it's a sufficient way to output information.
I like the variable arguments count option a lot, didn't even think about it.
I'll make sure it's in the rewriting and then I'll work out a better implementation for the Classes I'll work on including ArduinoIoTCloud :)

PR #33, opened 2019-02-13
Add debug print for broker host string

endorama:
One of possible reasons for not being able to connect to the IoT Cloud could be trying to connect to the wrong broker address.

This is particularly true in testing environments.

Adding a low priority debug message showing the broker connection string just before connecting can help debug this kind of issues.


ubidefeo:
already implemented in #34

PR #34, opened 2019-02-14
print troubleshooting info, added broker port, general cleanup

ubidefeo:
ArduinoIoTCloudClass:
- defined DEFAULT_BROKER_PORT (see next)
- added brokerPort private member (uint16_t because of how it is defined in MQTT client)
- added possibility to pass an alternative Broker port after the address
- added getBrokerPort() method
- added printDebugInfo() method to ArduinoIoTCloud, to allow user to print out info useful for troubleshooting: Broker address:port, Thing ID, Device ID
WiFiConnectionManager:
- tweaked interval time for GETTIME
- added check for firmwareVersion() required version (delays 5 seconds if version is older than latest fixed ones, but still moves forward)
- cleaned up a couple of debugMessage formatted strings
Both:
removed *msgBuffer=0 from anywhere the buffer was filled with a new message

### DIFF
--- a/src/WiFiConnectionManager.h
+++ b/src/WiFiConnectionManager.h
@@ -33,7 +33,7 @@ private:
 
   const int CHECK_INTERVAL_INIT = 100;
   const int CHECK_INTERVAL_CONNECTING = 500;
-  const int CHECK_INTERVAL_GETTIME = 100;
+  const int CHECK_INTERVAL_GETTIME = 666;
   const int CHECK_INTERVAL_CONNECTED = 10000;
   const int CHECK_INTERVAL_RETRYING = 5000;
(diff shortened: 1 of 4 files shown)


src/WiFiConnectionManager.h
```diff
@@ -33,7 +33,7 @@ class WiFiConnectionManager : public ConnectionManager {
 
   const int CHECK_INTERVAL_INIT = 100;
   const int CHECK_INTERVAL_CONNECTING = 500;
-  const int CHECK_INTERVAL_GETTIME = 100;
+  const int CHECK_INTERVAL_GETTIME = 666;
```
lxrobotics: Any specific reason the value is exactly 666? ;)

PR #35, opened 2019-02-15
possibility to disable logging

ubidefeo:
- `setDebugMessageLevel()` will now accept -1 to disable logging

### DIFF
--- a/src/ConnectionManager.h
+++ b/src/ConnectionManager.h
@@ -18,9 +18,7 @@
 #ifndef CONNECTION_MANAGER_H_INCLUDED
 #define CONNECTION_MANAGER_H_INCLUDED
 
-#ifndef ARDUINO_CLOUD_DEBUG_LEVEL
 #define ARDUINO_CLOUD_DEBUG_LEVEL 2
-#endif
 
 #include <Client.h>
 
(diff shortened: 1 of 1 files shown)


src/ConnectionManager.h
```diff
@@ -18,9 +18,7 @@
 #ifndef CONNECTION_MANAGER_H_INCLUDED
 #define CONNECTION_MANAGER_H_INCLUDED
 
-#ifndef ARDUINO_CLOUD_DEBUG_LEVEL
```
facchinm: I'd leave the guard to allow multiple inclusion of this file

src/ConnectionManager.h
```diff
@@ -77,7 +75,10 @@ class ConnectionManager {
 #endif
 
 static int debugMessageLevel = ARDUINO_CLOUD_DEBUG_LEVEL;
-inline void debugMessage(char *_msg, uint8_t _debugLevel, bool _timestamp = true, bool _newline = true) {
+inline void debugMessage(char *_msg, int _debugLevel, bool _timestamp = true, bool _newline = true) {
+  if(_debugLevel < 0){
```
facchinm: this check is probably not needed if we only allow messages from 0 up :slightly_smiling_face:

src/ConnectionManager.h
ubidefeo: @facchinm this is only to define a default in case it wasn't defined in the sketch, but it's definitely not needed anymore.
there's still a check against multiple inclusion

src/ConnectionManager.h
ubidefeo: mh... better safe than sorry? :)

PR #37, opened 2019-02-22
Managed exit cases for IOT_STATUS_CLOUD_CONNECTED, removed unused states

ilcato:
It manages the case in which despite having the underlying network connection the board loses the connection to the mqtt broker. Without this change it would remain locked indefinitely into this state until restart.


ilcato:
Already implemented in #31

PR #45, opened 2019-03-26
change thing id to device id

AlbyIanna:
The id we are storing on the crypto chip is actually the **device** id, it has nothing to do with a **thing**.


examples/utility/Provisioning/Provisioning.ino
```diff
@@ -9,7 +9,7 @@ const bool DEBUG = true;
 const int keySlot                                   = 0;
 const int compressedCertSlot                        = 10;
 const int serialNumberAndAuthorityKeyIdentifierSlot = 11;
-const int thingIdSlot                               = 12;
+const int deviceIdSlot                               = 12;
```
lxrobotics: Please keep the formatting 😉 
```suggestion
const int deviceIdSlot                              = 12;
```

PR #60, opened 2019-04-23
Simplifying getting started examples ...

lxrobotics:
by extracting the arduino_secrets into separate file.

@ubidefeo @AlbyIanna @umbynos I further suggest to rename the examples into something which contains *getting started* as a reminder of what's the purpose of those examples, e.g. `GSM_Cloud_Blink` -> `MKR_GSM_1400_Getting_Started` or `MKR_GSM_1400_Getting_Started_Cloud_Blink`. What's your opinion?

### DIFF
--- a/examples/ArduinoIoTCloud_LED_switch/ArduinoIoTCloud_LED_switch.ino
+++ b/examples/ArduinoIoTCloud_LED_switch/ArduinoIoTCloud_LED_switch.ino
@@ -10,6 +10,10 @@
   By default, settings for WiFi are chosen. If you prefer to use a GSM board take a look at thingProperties.h arduino_secrets.h,
   to make sure you uncomment what's needed and comment incompatible instructions.
 
+  This sketch is compatible with:
+   - MKR 1000
+   - MKR WIFI 1010
+   - MKR GSM 1400
 */
 #include "arduino_secrets.h"
 #include "thingProperties.h"
--- a/examples/ArduinoIoTCloud_Travis_CI/ArduinoIoTCloud_Travis_CI.ino
+++ b/examples/ArduinoIoTCloud_Travis_CI/ArduinoIoTCloud_Travis_CI.ino
@@ -2,6 +2,11 @@
    This sketch is used in combination with Travis CI to check if
    unintentional breaking changes are made to the used facing
    Arduino IoT Cloud API.
+
+   This sketch is compatible with:
+     - MKR 1000
+     - MKR WIFI 1010
+     - MKR GSM 1400
 */
 
 #include "arduino_secrets.h"
--- a/examples/GSM_Cloud_Blink/GSM_Cloud_Blink.ino
+++ b/examples/GSM_Cloud_Blink/GSM_Cloud_Blink.ino
@@ -2,25 +2,7 @@
 #include <ConnectionManager.h>
 #include <GSMConnectionManager.h>
 
-#ifndef SECRET_PIN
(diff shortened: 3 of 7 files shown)


umbynos:
Originally I named the examples that way because of [this](https://github.com/bcmi-labs/create-gettingstarted-site/blob/master/src/app/gettingstarted/sketches/MKRGSM1400_Cloud_Blink.ino.js). But I think is good to rename them with something more specific.

lxrobotics:
@umbynos @ubidefeo @AlbyIanna I've renamed the sketches yesterday. Is there anything else in the way of merging this PR?

AlbyIanna:
@lxrobotics I'm dubious about the renaming. I don't think we need to have the label `Getting_Started` in the name, it feels confusing to me. Also, why do we need to specify the name of the boards?

lxrobotics:
@AlbyIanna 
> I don't think we need to have the label Getting_Started in the name, it feels confusing to me.

I disagree, it makes a lot of sense to me to name them just so. The other examples within the example folder are intended to be consumed by users of the `ArduinoIoTCloud` library. The "Getting Started" sketches are only here so we can Travis-CI verify their integrity with the `ArduinoIoTCloud` library while their main purpose is to be used within the getting started procedure (performed via `Arduino Create`).

> Also, why do we need to specify the name of the boards?

Without having the target board within the filename you can't know without looking into the code which for which board the example is provided. Having the target board name encoded in the filename removes the need for looking into the code. 

Otherwise there is a real possibility of uploading the wrong sketch to the wrong board which will result in the sketch not working.

We could rewrite the code in such a way that there would be only a single sketch for all target boards but this would involve more code and some ifdefs which I gather from your previous voiced opinion (on slack) is not what you want.

lxrobotics:
@AlbyIanna @umbynos Since nearly a month has passed I'd like to re-request your opinion on the subject. **No rename** is also fine by me but I'd like to have clarity on the subject.
@ubidefeo @mattiabertorello Do you have a opinion on the subject?

AlbyIanna:
Hi @lxrobotics, sorry for the late response.

> I disagree, it makes a lot of sense to me to name them just so. The other examples within the example folder are intended to be consumed by users of the ArduinoIoTCloud library. The "Getting Started" sketches are only here so we can Travis-CI verify their integrity with the ArduinoIoTCloud library while their main purpose is to be used within the getting started procedure (performed via Arduino Create).

Actually those sketches are not there to the only purpose of being used by the Getting Started, but instead they are full-fledged examples: they show you how to connect to the Cloud and use the Cloud Monitor. IMO the fact that we are using them in the GS is collateral.


> Without having the target board within the filename you can't know without looking into the code which for which board the example is provided. Having the target board name encoded in the filename removes the need for looking into the code.
> Otherwise there is a real possibility of uploading the wrong sketch to the wrong board which will result in the sketch not working.

Good point, but as we keep adding new GS flows for other boards (for example, the Uno WiFi and the Nano IoT) we cannot keep to add them to the example name.

A solution @mattiabertorello and I both agreed with is to add a comment at the beginning of the sketch, specifying the compatible boards (and maybe a mention to the fact that we are using this sketch in the GS, and a link).

lxrobotics:
Hi @AlbyIanna and @mattiabertorello - thank you very much for the feedback :smiley: I've implemented your proposed changes, please have a look at them and let me know if they are what you had in mind.

AlbyIanna:
LGTM!

PR #62, opened 2019-04-30
Multi value properties management

ilcato:
This PR implement phase 1 and 2 of the multi value property concept from the client/device perspective.
It is fully compatible with the previous versions and allows for a mix of monovalue properties and multi value properties.
Monovalues properties can be expressed with either a primitive type or an equivalent object type, e.g.:
`bool Switch;`
or
`CloudBool Switch`
by using the latter format there's a better precision compute of locally changes timestamps and the opportunity to have a more responsive cloud alignment independent by the sketch code (not yet implemented).

This is also compatible with Phase 3; a formalization of the steps that explain how to create a new property type definition is required.


ubidefeo:
@ilcato I guess timestamp is handled on a property level, as I see the method gone from this update.
it looks good, a bit out of my understanding when it comes to some parts of the code.
I can try it out tomorrow to see how behaves.
@lxrobotics what's your take?

ilcato:
@ubidefeo, local timestamp measurement is now:

- for primitive types properties in `Thing.updateTimestampOnLocallyChangedProperties();`

- for object based (both mono and multi value) properties in operator= overload of the corresponding class.

lxrobotics:
Successfully tested with `MKR1000`.

PR #64, opened 2019-05-08
Multi value properties management

ilcato:
This PR implement phase 1 and 2 of the multi value property concept from the client/device perspective.
It is fully compatible with the previous versions and allows for a mix of monovalue properties and multi value properties.
Monovalues properties can be expressed with either a primitive type or an equivalent object type, e.g.:
bool Switch;
or
CloudBool Switch
by using the latter format there's a better precision compute of locally changes timestamps and the opportunity to have a more responsive cloud alignment independent by the sketch code (not yet implemented).

This is also compatible with Phase 3; a formalization of the steps that explain how to create a new property type definition is required.


lxrobotics:
@ilcato I've checked the Travis CI build log and I see that the code formatting check fails as well - https://travis-ci.org/arduino-libraries/ArduinoIoTCloud/jobs/529875154 . While its okay for the example build to fail because of the dependency to `ArduinoCloudThing` there is no such reason for the code formatting check. Please fix this.

PR #65, opened 2019-05-15
Simplify multi value properties sketch

mattiabertorello:
Due to the feedback received, it necessary to change the multi value properties in some way more compliance to the Arduino style. 
This type of initialization can be a solution
https://processing.org/reference/PVector.html
`location  = Location(latMov, lonMov);`
instead 
`Loc =   { .lat = latMov,   .lon = lonMov };`


lxrobotics:
@mattiabertorello I've just noticed that this example has not been added to `.travis.yml` and is as a result not covered by the CI build. Please add `MultiValue_example.ino` to travis first.

Issue #67, opened 2019-05-17
Fix c++ warnings


cmaglie:
After merging #66 and looking at the Travis build logs, it looks like this library triggers many c++ warnings.
Even if most of them are innocuos I think that it's worth trying to solve them anyway since sometimes they reveal some very hidden bugs.

lxrobotics:
Fixed by #69.

PR #69, opened 2019-05-21
Fixing compiler warnings due to stronger compile-time checks

lxrobotics:
With regards to #67 .

### DIFF
--- a/examples/ArduinoIoTCloud_LED_switch/arduino_secrets.h
+++ b/examples/ArduinoIoTCloud_LED_switch/arduino_secrets.h
@@ -1,12 +1,12 @@
 #include <ConnectionManager.h>
 
-/* (MKR1000, MKR WiFi 1010) */
+/* MKR1000, MKR WiFi 1010 */
 #if defined(BOARD_HAS_WIFI)
   #define SECRET_SSID "YOUR_WIFI_NETWORK_NAME"
   #define SECRET_PASS "YOUR_WIFI_PASSWORD"
 #endif
 
-/* MKR GSM 1400 comment the lines above and uncommet the following. */
+/* MKR GSM 1400 */
 #if defined(BOARD_HAS_GSM)
   #define SECRET_PIN ""
   #define SECRET_APN ""
--- a/examples/ArduinoIoTCloud_LED_switch/arduino_secrets.h
+++ b/examples/ArduinoIoTCloud_LED_switch/arduino_secrets.h
@@ -1,15 +1,9 @@
-/*
-  Fill in your login credentials:
+/* (MKR1000, MKR WiFi 1010) */
+#define WIFI_SECRET_SSID "YOUR_WIFI_NETWORK_NAME"
+#define WIFI_SECRET_PASS "YOUR_WIFI_PASSWORD"
 
-  The following lines are used for WiFi enabled boards (MKR1000, MKR WiFi 1010)
-*/
-#define SECRET_SSID "YOUR_WIFI_NETWORK_NAME"
-#define SECRET_PASS "YOUR_WIFI_PASSWORD"
-/*
-  If you prefer using a MKR GSM 1400 comment the lines above and uncommet the following.
-  PIN, APN, Login and Password are supplied by your Cellular Data provider.
-*/
-#define SECRET_PIN ""
-#define SECRET_APN ""
-#define SECRET_LOGIN ""
-#define SECRET_PASS ""
+/* MKR GSM 1400 comment the lines above and uncommet the following. */
+#define GSM_SECRET_PIN ""
+#define GSM_SECRET_APN ""
+#define GSM_SECRET_LOGIN ""
+#define GSM_SECRET_PASS ""
--- a/examples/ArduinoIoTCloud_LED_switch/arduino_secrets.h
+++ b/examples/ArduinoIoTCloud_LED_switch/arduino_secrets.h
@@ -1,9 +1,15 @@
+#include <ConnectionManager.h>
+
 /* (MKR1000, MKR WiFi 1010) */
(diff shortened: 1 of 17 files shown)


examples/ArduinoIoTCloud_LED_switch/arduino_secrets.h
```diff
@@ -1,15 +1,9 @@
...
-#define SECRET_PASS "YOUR_WIFI_PASSWORD"
-/*
-  If you prefer using a MKR GSM 1400 comment the lines above and uncommet the following.
-  PIN, APN, Login and Password are supplied by your Cellular Data provider.
-*/
-#define SECRET_PIN ""
-#define SECRET_APN ""
-#define SECRET_LOGIN ""
-#define SECRET_PASS ""
+/* MKR GSM 1400 comment the lines above and uncommet the following. */
```
ubidefeo: typo on "uncommet"

src/ArduinoIoTCloud.cpp
```diff
@@ -418,6 +412,7 @@ void ArduinoIoTCloudClass::connectionCheck() {
 
 void ArduinoIoTCloudClass::setIoTConnectionState(ArduinoIoTConnectionStatus _newState) {
```
ubidefeo: just a comment on style:
if we use an underscore prefix to characterise member variables we should not use them in signatures.
I have been always used to characterise with underscore my signature ones, and it's very possible I wrote this method in the first place.
let's pick one, but since I've seen underscores in all the member properties we should change this one

src/EthernetConnectionManager.cpp
```diff
@@ -128,13 +128,21 @@ void EthConnectionManager::check() {
           debugMessage(DebugLevel::Info, "Connected");
         }
         break;
+      case NetworkConnectionState::DISCONNECTING: {
```
ubidefeo: we added these ones back in?
I know they do nothing but I have always like them

src/GSMConnectionManager.cpp
```diff
@@ -106,11 +105,23 @@ void GSMConnectionManager::check() {
           debugMessage(DebugLevel::Verbose, "Connected to Cellular Network");
         }
         break;
+      case NetworkConnectionState::GETTIME: {
```
ubidefeo: I believe these were taken out by @ilcato to optimise.
may I ask why they have been reintroduced?
I like them there for future use

src/WiFiConnectionManager.h
```diff
@@ -52,19 +52,19 @@ class WiFiConnectionManager : public ConnectionManager {
...
     void changeConnectionState(NetworkConnectionState _newState);
 
-    const int CHECK_INTERVAL_IDLE = 100;
-    const int CHECK_INTERVAL_INIT = 100;
-    const int CHECK_INTERVAL_CONNECTING = 500;
-    const int CHECK_INTERVAL_CONNECTED = 10000;
-    const int CHECK_INTERVAL_RETRYING = 5000;
-    const int CHECK_INTERVAL_DISCONNECTED = 1000;
-    const int CHECK_INTERVAL_ERROR = 500;
+    const unsigned long CHECK_INTERVAL_IDLE = 100;
```
ubidefeo: since these are just intervals of milliseconds never exceeding the 60 seconds, aren't we just fine wth `uint16_t`?
this is valid for all of the ones coming afterwards

examples/ArduinoIoTCloud_Travis_CI/arduino_secrets.h
```diff
@@ -1,9 +1,9 @@
 /* (MKR1000, MKR WiFi 1010) */
-#define SECRET_SSID "YOUR_WIFI_NETWORK_NAME"
-#define SECRET_PASS "YOUR_WIFI_PASSWORD"
+#define WIFI_SECRET_SSID "YOUR_WIFI_NETWORK_NAME"
```
mastrolinux: This is not going to work, the parser looks for SECRET string as being the first part of the word.
Then please rename SECRET_WIFI_SSID and so on.

src/EthernetConnectionManager.cpp
lxrobotics: When there is a switch statement evaluating a enum the compiler can determine wether or not all possible value cases are covered within the switch statement (`-Wswitch-enum`). If one or more of those are not found the compiler issues a warning.

Usually missing cases indicate a programming problem because usually when one implements a switch-case with a enumeration type it is for implementing *finite state machines* (which is actually kind of what we do in the code). I've therefore added the empty if statements as a reminder for the programmer that there are states which are not covered and might present a possible problem (either in the future or already now).

src/GSMConnectionManager.cpp
lxrobotics: Same reason for reintroducing them as in the statement above - additionally I'd like to quote Donald Knuth who said **"Premature optimization is the root of all evil"** :smile: We are not in High-Frequency-Trading, clarity and expressiveness of code should matter most.

src/WiFiConnectionManager.h
lxrobotics: The return type of `millis` is of type `unsigned long` - if we reduce those constants to `uint16_t` we would need to cast at the point of their usage.

examples/ArduinoIoTCloud_Travis_CI/arduino_secrets.h
lxrobotics: Done :smile:

examples/ArduinoIoTCloud_LED_switch/arduino_secrets.h
lxrobotics: Fixed ;)

lxrobotics:
Successfully checked with `MKR1000` - @mattiabertorello or @ubidefeo can you please verify with a `MKR GSM 1400`?

PR #72, opened 2019-05-24
Reducing available update functions

lxrobotics:
In 20cbb75e70d0360138189c5a8cf6992942ce1247 the implementation of

`ArduinoIoTCloudClass::update(int const reconnectionMaxRetries, int const reconnectionTimeoutMs)` 

was changed in such a way, that the function parameters `reconnectionMaxRetries` and `reconnectionTimeoutMs` ceased to have any meaning since they were no longer used. This fact was uncovered when enforcing stronger compile time checks in f43c0fb7b19c179be5b9f49ee6acacbd58f83ed3.

Capitalizing on this discovery we can drastically reduce the number of overloaded `update` functions provided. We may even remove the functions marked `deprecated` at a later point in time and we are left with a single `update` function.

### DIFF
--- a/src/ArduinoIoTCloud.cpp
+++ b/src/ArduinoIoTCloud.cpp
@@ -211,7 +211,7 @@ bool ArduinoIoTCloudClass::disconnect() {
   return true;
 }
 
-void ArduinoIoTCloudClass::update(int const /* reconnectionMaxRetries */, int const /* reconnectionTimeoutMs */, CallbackFunc onSyncCompleteCallback) {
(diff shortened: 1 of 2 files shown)


ubidefeo:
I think we should have a way to limit retries at some point.
some use cases in low power contexts or moving setups might involve devices which wake up, try to connect and go back to sleep if unsuccessful.
I agree that not handling the cases is silly, so we maybe should

lxrobotics:
I agree the reconnection attempts should be limited, however, this PR is about eliminating dead parameters. If we actually have a reconnection strategy in some future point in time we may reintroduce a update function where one can configure the reconnection parameters.

PR #74, opened 2019-06-17
add nano iot connection support [WIP]

AlbyIanna:


ubidefeo:
@AlbyIanna , I think the build fails because it needs a newer core/toolchain to compile.
@facchinm can you confirm?

facchinm:
@ubidefeo it fails due to
`/home/travis/Arduino/libraries/MKRGSM/src/GSM.cpp:340:7: error: 'strptime' was not declared in this scope
`
which is in fact fixed by _XOPEN_SOURCE patch

```patch
   You should have received a copy of the GNU Lesser General Public
   License along with this library; if not, write to the Free Software
   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#define _XOPEN_SOURCE
 #include <time.h>
 
 #include "Modem.h"
 
 #include "NB.h"
```
Would you mind filing the PR on both GSM and NB repos? :kissing:

ubidefeo:
@facchinm you had me apply this fix a while back, which is, in fact, in my local source and I had completely forgotten about it.
I'll do the PR for both libs this morning 🙏

AlbyIanna:
:cry:

AlbyIanna:
:smile:

ubidefeo:
@AlbyIanna I have changed the title of this issue not to let people think this enables support for the Nano 33 IoT as it is, in fact, a work in progress towards that end :)

PR #76, opened 2019-08-08
ESP8266 support for Arduino IoT Cloud (User/Pass Authentification)

lxrobotics:
This pull request adds support for the ESP8266 to be used with the [Arduino IoT Cloud](https://create.arduino.cc/iot) utilizing **user/password authentification** with the cloud.

If you want to try this feature out you must follow the following steps:
* Checkout the corresponding branch of `Arduino_ConnectionHandler` which contains the necessary changes to support ESP8266 user/pass login `git clone https://github.com/arduino-libraries/Arduino_ConnectionHandler -b esp8266-user-pass-login`
* A example sketch is available under `examples/ArduinoIoTCloud_ESP8266`
  * Set your `ARDUINO_CLOUD_DEVICE_ID`, `ARDUINO_CLOUD_DEVICE_PASS` and `THING_ID` as well as your WIFI data `SECRET_SSID` and `SECRET_PASS`

Requires https://github.com/arduino-libraries/Arduino_ConnectionHandler/pull/1


lxrobotics:
I don't know how the codacy/PR quality review was added to the CI build (maybe @per1234 or @gvarisco can shed some light on that) but I strongly suggest to remove it. I've got no intention of working through the useless formatting suggestions for `README.md`, with the value of those suggestions nearing zero.

gvarisco:
> I don't know how the codacy/PR quality review was added to the CI build (maybe @per1234 or @gvarisco can shed some light on that) but I strongly suggest to remove it. I've got no intention of working through the useless formatting suggestions for `README.md`, with the value of those suggestions nearing zero.

Removed it.

lxrobotics:
Branch successfully tested with **ESP8266** based board and **MKR1000**. @ubidefeo Can you please try it out with a **MKRGSM1400**?

eclipse1985:
> Branch successfully tested with **ESP8266** based board and **MKR1000**. @ubidefeo Can you please try it out with a **MKRGSM1400**?

Just tested with a GSM 1400, it works

AlbyIanna:
I tested the example sketches used in the Getting Started and it worked very well, but there is just a little problem on my side: the message printed on the serial monitor when the board connects to the cloud has changed from `"Connected to Arduino IoT Cloud"` to `"Arduino IoT Cloud Connection status: CONNECTED"`. This caused the Getting Started to fail on the first try. It's not a big issue to me to change a string, obviously, but I'd really like to make sure that when an important string like that changes, I get informed so that I can change it accordingly on the front-end. Maybe there are other strings that I missed.
A simple solution could be to add some tests to check that the strings I rely don't change, so that when they do, you would that the front-end need to change accordingly.
@lxrobotics what do you think?

AlbyIanna:
Here is a list of the strings I'm checking during the Getting Started:
- `WiFi Hardware failure`
- `GPRS attach failed`
- `No Ethernet chip connected`
- `Connected to Arduino IoT Cloud`
- `Connected to`, the first time I receive this string, to me it means that the board successfully connected to the network (GSM or WiFi)

lxrobotics:
Hi @AlbyIanna :wave: sorry for messing with your string evaluation, I was not aware you are using that to determine system state. I've checked the strings you've provided in the above comment against the current implementation:
- `"WiFi Hardware failure"` :heavy_check_mark: 
- `"GPRS attach failed"` :heavy_check_mark: 
- `"No Ethernet chip connected"` :x: - Currently there is no support for boards connected via Ethernet, therefore there is no such error message. If that should change we can of course reintroduce this exact error message.
- `"Connected to Arduino IoT Cloud"` :x: As you already found out this message reads now `"Arduino IoT Cloud Connection status: CONNECTED"`.

lxrobotics:
Rebased the changes of https://github.com/arduino-libraries/ArduinoIoTCloud/tree/arduinobearssl on top of this branch, therefore eliminating the need to maintain 2 different branches what will become one release in any case.

PR #78, opened 2019-08-20
Replace ArduinoIoTCloudBearSSL with ArduinoBearSSL

lxrobotics:


ubidefeo:
Amazing job, @lxrobotics 🙏

lxrobotics:
@ubidefeo - if only it would work. There seem to be some subtle differences between the two libraries -  investigating.

lxrobotics:
@ubidefeo It works now :wink:

ubidefeo:
must have been that early morning clarity 👌🏼

PR #79, opened 2019-09-27
Add new cloud blink sketch version working with security credentials

AlbyIanna:
We need to had a new example sketch for the Getting Started flow for ESP8266.
Also, since the `DEVICE_LOGIN_NAME` was actually the board id, I changed it to `BOARD_ID`, which I believe makes more sense for the users.

### DIFF
--- a/examples/ArduinoIoTCloud_ESP8266/thingProperties.h
+++ b/examples/ArduinoIoTCloud_ESP8266/thingProperties.h
@@ -2,7 +2,7 @@
 #include <Arduino_ConnectionHandler.h>
 
 #define THING_ID "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" /* "Thing ID" when selecting thing within Arduino Create */
-#define DEVICE_LOGIN_NAME "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+#define BOARD_ID "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
 
 void onLedChange();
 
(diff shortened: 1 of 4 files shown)


lxrobotics:
Hi @AlbyIanna :wave: your PR looks good to me 👍 However, I've chosen `DEVICE_LOGIN_NAME` on request of @mastrolinux since that was what the ID was called within a design document (for which I unfortunately can't find the link right now). @mastrolinux I'd agree with @AlbyIanna that `BOARD_ID` makes more sense for the user since this is what he sees when he is within the Create IDE. What's your final word on this?

mastrolinux:
Hi @lxrobotics indeed after the last changes in the UI and in the backend I think it is OK to use BOARD_ID. It is not going to change in the short term and we will not have a different name than the board id. 

Thank you

mastrolinux:
I do not get why we have SECRET_BOARD_ID and BOARD_ID at the same time. @AlbyIanna

AlbyIanna:
@mastrolinux They are two different examples:
- https://github.com/arduino-libraries/ArduinoIoTCloud/pull/79/files#diff-36a58c194edbcf0742389ac295e372a1 - since this is an example of a cloud sketch, I assumed that we could automatically set the `BOARD_ID` (exactly like what we do with the `THING_ID`), no reason to make it `SECRET`, it would just require extra work from the user, who would need to change it manually whenever the board changes.
- https://github.com/arduino-libraries/ArduinoIoTCloud/pull/79/files#diff-ad8dc2c2de15d6f35c8dc59a55c45e9a - instead, this one is the example used in the Getting Started, which is used only once, and is intended to show the user that we need to use the credentials they just generated (we are going to auto-fill the form anyway)

PR #90, opened 2020-02-21
[WIP] Remove no longer necessary module CloudSerial

aentinger:
The class `CloudSerial` was initally introduced to implement the getting started workflow for the various cloud enabled boards. Since the cloud capabilities have progressed since we can also realise getting started via properties which is why there is no longer a need for keeping the class `CloudSerial` in the code.


aentinger:
Closing this PR since we might need `CloudSerial` after all, but it would make a lot of sense to move it to a 3rd party library.

PR #93, opened 2020-02-26
Register a epoch timestamp retrieval function with ArduinoCloudThing

aentinger:
This allows class ArduinoCloudThing to obtain a 'epoch' timestamp which is used during synchronisation and which has been implemented so far via an 'extern' dependency to RTCZero. Since this kind of hidden dependency (basically a global object across 2 separate Arduino libraries) is very brittle the depdency is made clear by explicitly registering a function to provide the required timestamp.


aentinger:
Successfully tested on **PROD** with

PR #94, opened 2020-02-26
Removal of references to Client class since it's not used anymore.

aentinger:
Cleaning up by removing pointer to a Client class which does not make sense any longer because we always use the connection handler to access any kind of connection


aentinger:
Successfully tested with

PR #110, opened 2020-04-06
Cleaning up examples

aentinger:


aentinger:
Hi @AlbyIanna :wave: Since I suppose the getting started code used on Create is also not directly taken from this repository it's okay to move the cloud blink examples to the subfolder `utility`?

Issue #73, opened 2019-05-27
add a ArduinoIoTCloudClass::end() function for a regulated shutdown


generationmake:
I suggest to add a ArduinoIoTCloudClass::end() function for a regulated shutdown and restart of the whole cloud function. This would allow arduinos to shut down the cloud functions for power reasons (for example sleep mode) and enable them later again.

For example:
```
void ArduinoIoTCloudClass::end() {
  _connection->end();  // power down network interface - reduces power consumption from serveral mAs to a few uAs
  ECCX08.end();     // power down ECC508 - reduces power consumption from 800 uA to 6 uA
}

```
and add to the ConnectionManager a public end function. For the MKR WIFI 1010 for example WiFi.end();

Thank you

ubidefeo:
hi @generationmake 
You have a really good point there.
We have been working on an update which will allow better control of connection/disconnection.
Of course our goal is to work on something which won't introduce breaking changes, but we're getting there.
Stay tuned :)

lxrobotics:
Hi @generationmake 
Thank you very much for raising this issue. If you have already a complete implementation feel free to create a pull request. I'll have to advise you though that currently refactoring efforts are under way which might render your PR obsolete, so I suggest the best course of action is to wait it out a bit and then create a PR with your suggested changes.

generationmake:
Hi @ubidefeo and @lxrobotics.
Thank you for your answers.
I will stay patient and see what comes up next.

podhrmic:
Hi! What is the status of this feature?
I have an application that needs to sleep for extended periods of time between updates (~10min), and I am not sure if the Arduino can reconnect with the cloud after the deep sleep (using Arduino Low Power library on MKR GSM board).
Is that currently possible?

ubidefeo:
hi @podhrmic 
I have successfully used the automatic reconnect on IoT Cloud using MKR GSM waking up from deep sleep.
I have occasionally had reconnection issues when coverage wasn't available but it worked once signal was fine again

podhrmic:
Thanks for the quick reply!

ubidefeo:
my pleasure, @podhrmic 
just make sure you give the board enough time to reconnect before it goes back to sleep ;)

andreacosti:
hi @ubidefeo 
I have the same problem. I have a MKRGSM1400 with a MKRENV shield and I want to send dato to cloud.
[...]

PR #113, opened 2020-04-08
Reduce public API by turning connect/disconnect into protected methods.

aentinger:
The public `ArduinoIoTCloud` member methods `connect` and `disconnect` should not be called by the user of the library (they are called internally) and are therefore transformed to `protected` member functions. In the future a `end` function should be added (see #73) which works the opposite way to the the `begin` method and cleans everything up.


aentinger:
* `MKR WAN 1310` :heavy_check_mark:

PR #128, opened 2020-06-03
Both MqttClient and BearSSLClient are allocated on the stack instead of the heap.

aentinger:
This means that the memory consumption of those two rather large classes are directly visibale in the compilation output since ArduinoIoTCloud is instantiated as a global variable. The heap allocation is not directly visible to the user of the library and so gives the wrong impression that there is a lot of free space left.


aentinger:
* `MKR WAN 1310` :heavy_check_mark:

PR #132, opened 2020-06-04
Extracting property container functionality

aentinger:
So far the "container" functionality for holding all the properties has been exercised by the class ArduinoCloudThing which leads to this class doing business logic, cbor encoding/decoding and container services. In order to clean up the interfaces and make the code more understandable and extendable it is desirable to have only one purpose per class (SRP).


aentinger:

PR #135, opened 2020-06-08
Additional update policy "OnDemand"

aentinger:
This new update policy allows to trigger the transmission of all properties when a certain function (`push`) is called. In order to configure a property for this update policy it must be initalized using the following syntax in `thing_properties.h`:
```C++
int seconds;
/* ... */
void initProperties()
{
  ArduinoCloud.addProperty(seconds, Permission::Read).publishOnDemand();
```
In order to trigger the update the `ArduinoCloud.pushAll()` needs to be called. All properties are transmitted on the next call of `ArduinoCloud.update()`.
```C++
void loop()
{
  /* ... */
  ArduinoCloud.update();
  /* ... */
  ArduinoCloud.pushAll();
}
```

### DIFF
--- a/src/ArduinoIoTCloud.cpp
+++ b/src/ArduinoIoTCloud.cpp
@@ -25,6 +25,11 @@
  * PUBLIC MEMBER FUNCTIONS
  ******************************************************************************/
 
+void ArduinoIoTCloudClass::push()
+{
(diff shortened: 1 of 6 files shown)


aentinger:
Here's your update on demand policy @manchoz and @zmoog :wave:

PR #137, opened 2020-06-08
Allow an optional encoding of the timestamp in the CBOR message

aentinger:
In order to enable the encoding of the timestamp you can've got to call `encodeTimestamp()` when adding the property to the cloud within `thing_properties.h`.
```C++
ArduinoCloud.addProperty(seconds, Permission::Read).publishOnDemand().encodeTimestamp();
```
In order to set the timestamp of the property you call
```C++
seconds_timestamp = seconds; /* = 45 in below CBOR */
ArduinoCloud.setTimestamp("seconds", seconds_timestamp);
```
This is the resulting CBOR
```bash
[{0: "seconds", 2: 45, 6: 45}]
```
@zmoog :wave:


aentinger:

PR #152, opened 2020-07-06
Fix: Handle encoded CBOR message exceeding buffer size

aentinger:
Currently the payload buffer for CBOR encoded messages has a size of 256 bytes. If the encoded cbor message is exceeding this max payload size then the whole message is discarded. This PR should fix this problem by encoding properties until the payload buffer is full and encoding the remaining properties on the next run of `CBOREncoder::encode`.


aentinger:
@manchoz this might be for you.

aentinger:

PR #160, opened 2020-07-08
Fix: Use sketch compile time if no connection is available for estimating system time.

aentinger:
Ensure that the returned time is at least the compile time of the sketch, otherwise we might have an error scenario where no connection is available which would lead to a timestamp of 0 which would lead to the certificate being declared 'expired' and the device will be unable to connect to the server


aentinger:
Successfully tested on DEV with MRK WiFi 1010 :heavy_check_mark:

PR #190, opened 2020-08-21
Removing stdin/stdout topic strings ...

aentinger:
since we currently don't need those topics anymore (we might need them again once we reintegrate cloud serial but right now they just consume memory.


aentinger:
Tested with `MKR WiFi 1010` :heavy_check_mark:

PR #202, opened 2020-10-02
OTA Header set compression flag

aentinger:
CC @eclipse1985 @zmoog

@zmoog Does this produce the same binary output as your builder extension?


aentinger:
builder and python tools result in same OTA file as tested by @aentinger .

PR #205, opened 2020-10-22
Fix lenght typo

eclipse1985:


per1234 (commented):
I'm glad to see this fix! I seem to always make this typo myself; maybe something to do with the keyboard layout?

My only concern would be that this could be a breaking change if people are using the `CloudLenght` type in their code. I see there have been several releases of the library with it in place. I don't know enough about this to say whether that's at all likely. If it is possible that people are using it, the misspelled type could be left in place for backwards compatibility and a new typedef added with the correct spelling.

eclipse1985:
Actually, this is a new kind of property, still not used in production as now, from what I see in the database.
The previous type `LENGHT` will be deprecated.

per1234 (approved):

PR #212, opened 2020-12-03
Adding alias for function onUpdate to address issues uncovered during user testing

aentinger:
Now functions which are triggered on an external change of the property, that is via the cloud dashboard can be registered via both `onUpdate` as well as `onExternalChange`.


aentinger:
Closing this PR due to a misunderstanding, the user confusion originated from the auto-generated names of the callbacks registered with this function, not from this function itself.

PR #231, opened 2021-02-21
fix typo in the ArduinoIoTCloud-Callbacks example

PhysicsUofRAUI:
I found an error in the Callbacks example when I was looking for a solution to a problem I was having with a project. It was the callback for what to do on disconnect. 

It was written:

```
ArduinoCloud.addCallback(ArduinoIoTCloudEvent::CONNECT, doThisOnDisconnect);
```

And I changed it to:

```
ArduinoCloud.addCallback(ArduinoIoTCloudEvent::DISCONNECT, doThisOnDisconnect);
```

Which I think is what it was meant to be.

This is my first time contributing to any Arduino project, so I apologize for breaking any conventions or not following proper protocol.


aentinger (approved):
LGTM :+1: Thank you @PhysicsUofRAUI :rocket:

Issue #221, opened 2021-01-13
ArduinoCloud does not reconnect


jcham:
The simple example projects to light up an LED work fine, but the whole thing stops working when WiFi connection gets momentarily lost, and it doesn't reconnect, requiring a reset.
If this is normal, then what is the recommended way to detect a disconnect and reconnect?
Thanks

jcham:
I've tried several things, but none of them cause the connection to come back up:

```
void loop() {
  
  ArduinoCloud.update();

  // ---- Maintaining connection led
  bool connected = ArduinoCloud.connected();
  if (connected_prev != connected) {
    Serial.print("ArduinoCloud connection ");
    Serial.println(connected);
  }
  connected_prev = connected;
  digitalWrite(CONNECTION_LED_PIN, connected);

  // ---- Reconnect
  ArduinoIoTPreferredConnection.check();
  // if (!connected) {
  //   Serial.println("ArduinoCloud reconnecting...");
  //   ArduinoIoTPreferredConnection.connect();
  //   ArduinoCloud.begin(ArduinoIoTPreferredConnection);
  //   delay(10000);
  // }
}
```

aentinger:
Hi @jcham :wave:
The reconnection is handled within the ArduinoCloud logic. Under no circumstances should you try to manually interfere with this logic (e.g. directly calling `ArduinoIoTPreferredConnection.check()`). When you say "WiFi connection gets momentarily lost" what kind of WiFi connection loss are you describing?

aentinger:
Also: What board are you using?

jcham:
Using an Arduino Nano 33. That's what I tried in the first place, but it never reconnects. I don't know in production why the connection fails, but I can replicate my problem by turning off the wifi at home. 
So not calling anything it should reconnect automatically?

jcham:
When I disconnect the WiFi, I get this on the Serial:
```
ArduinoIoTCloudTCP::handle_Connected MQTT client connection lost
```
But it never reconnects, what do I have to do to restart it.

jcham:
Ah I used the IDE to update the WiFi module firmware and now it's ok!

jcham:
Actually, there is still a problem: when the WiFi comes in and out it reconnects fine, but some other condition (after some random amount of time) causes it to drop and then it doesn't recover. How to debug this?

aentinger:
[...]

PR #233, opened 2021-03-01
Improve reliability on connection loss

### LINKED ISSUE #86: ArduinoIoTCloudClass::update() stuck when internet connection fails
Hi all,
I'm having trouble keeping a nano33iot running when the internet connection (but *not* wifi) fails/drops for a short period of time.
It gets stuck in the function `void ArduinoIoTCloudClass::update(CallbackFunc onSyncCompleteCallback)` called from the loop() code.

After the connection fails, the board will have 45/60 seconds of life before getting completely stuck inside that function. No trying to reconnect occurs.
Even if the internet connection fails for only a couple of seconds, the board will freeze in 45/60 seconds.
The only solution is to reset either the board or the access point (the latter will trigger a wifi drop and the board will come back to life again searching for the wifi network). 

Tested repeatedly in the last few days in many ways:
- unplugging the ethernet cable on a wifi accesspoint
- disabling the mobile network on android in hotspot mode
- casually walking away from the AP with the board in hands (this sometimes will trigger the wifi drop+reconnection)
- different accesspoint/networks
- on a mkrwifi1010

With some Serial.print-fu I was able to narrow it a little bit to the call `_mqttClient->poll()` inside the update() function. 

Tested sketch is fresh template generated by the Arduino Create IoT Cloud compiled with the latest release of ArduinoIoTCloud (and dependencies).

aentinger:
This fixes #86 and #221 (:pray:). @jcham @IgorSchelle can you please test?

CC @eclipse1985 @zmoog can please you test on your setup too?

### DIFF
--- a/src/ArduinoIoTCloudTCP.cpp
+++ b/src/ArduinoIoTCloudTCP.cpp
@@ -328,6 +328,14 @@ ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_ConnectMqttBroker()
 
 ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_SubscribeMqttTopics()
 {
+  if (!_mqttClient.connected())
+  {
+    DEBUG_ERROR("ArduinoIoTCloudTCP::%s MQTT client connection lost", __FUNCTION__);
+    _mqttClient.stop();
+    execCloudEventCallback(ArduinoIoTCloudEvent::DISCONNECT);
+    return State::ConnectPhy;
+  }
+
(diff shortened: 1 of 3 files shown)


eclipse1985:
Sure, I just uploaded a sketch that uses this PR.

IgorSchelle:
Hi, Of course.

po 1. 3. 2021 v 8:51 odesílatel Alexander Entinger <notifications@github.com>
napsal:

zmoog:
@aentinger deployed on my test weather station.

PR #239, opened 2021-04-19
Enable watchdog for ArduinoCore-mbed based boards

aentinger:


src/utility/watchdog/Watchdog.cpp
```diff
@@ -56,3 +73,31 @@ void wifi_nina_feed_watchdog()
   samd_watchdog_reset();
 }
 #endif /* ARDUINO_ARCH_SAMD */
+
+#ifdef ARDUINO_ARCH_MBED
+void mbed_watchdog_enable()
+{
+  watchdog_config_t cfg;
+#if defined(ARDUINO_PORTENTA_H7_M7) || defined(ARDUINO_PORTENTA_H7_M)
```
facchinm, line 81: `ARDUINO_PORTENTA_H7_M` should be `ARDUINO_PORTENTA_H7_M4` :wink:

src/utility/watchdog/Watchdog.h
```diff
@@ -18,21 +18,6 @@
...
- ******************************************************************************/
-
-#include <AIoTC_Config.h>
-
-/******************************************************************************
- * PREPROCESSOR SECTION
- ******************************************************************************/
-
-#ifdef ARDUINO_ARCH_SAMD
-#  include <Adafruit_SleepyDog.h>
```
facchinm, line 32: Is sleepydog included somewhere else?

facchinm (changes requested):

src/utility/watchdog/Watchdog.h
aentinger: No, that's the only spot. But this isn't a change particular to this PR, I simply moved the inclusion from .h to .cpp.

src/utility/watchdog/Watchdog.cpp
aentinger: Good catch!

PR #244, opened 2021-04-27
Implement exponential retry/back-off strategy to avoid router overloading.

aentinger:


facchinm:
I think the big overhead could be due to `pow()` or `std::min`. Can we use arduino's `min` or a simplified version of pow() based on bitshift?

aentinger:
Hi @facchinm :wave:
I was wondering too why such a large increase of flash consumption happened. Good catch! I've replaced it with bit-shift mechanics although I don't particularly like that due to readability. However, the reduction in flash speaks for itself :+1: 

With regards to `std::min` - that was a workaround to allow compilation on all platforms, I've also dropped it (the whole commit).

PR #246, opened 2021-04-28
Disable watchdog for MKR GSM 1400.

aentinger:
GSM connection establishment takes too long - watchdog times out.


aentinger:
Fixed via https://github.com/arduino-libraries/Arduino_ConnectionHandler/pull/58.

PR #251, opened 2021-05-06
enabled `setInsecure()` for ESP32

ubidefeo:
Previously a call to this method had to be omitted, it was only required for ESP8266.
Out of a debugging session ended up with the broker refusing connections so I enabled `insecure` mode and it now works


aentinger (approved):
The change looks reasonable for me, however I do not have any ESP32 hardware available to verify this change.

aentinger:
On a second thought its perfectly reasonable. If there is no secure element attached to the ESP32 then there's no other option available but to go the `setInsecure` (insecure) way. Please be advised @ubidefeo that using this method leads to a non-secure connection, as outlined in 6d0d5ed7dd9b7591b3e2ee2b103abf6711fd6c1c.

ubidefeo:
@aentinger 
sorry, didn't have time to get back to your last comment.
Previously using `setInsecure()` would fail compilation, that is why I only enabled it for ESP8266 in my first patch.
Now it's the other way around.
Thank you for merging :)

PR #277, opened 2021-09-29
Change CloudTime from CloudFloat to CloudUnsignedInt

pennam:
This pr adds a new `CloudUnsignedInt` type and changes `CloudTime` from `CloudFloat` to `CloudUnsignedInt`

/cc @eclipse1985

### DIFF
--- a/extras/test/src/test_decode.cpp
+++ b/extras/test/src/test_decode.cpp
@@ -382,6 +382,24 @@ SCENARIO("Arduino Cloud Properties are decoded", "[ArduinoCloudThing::decode]")
 
   /************************************************************************************/
 
+  WHEN("A Time property is changed via CBOR message")
+  {
+    PropertyContainer property_container;
+
+    CloudTime test;
+    test = 0;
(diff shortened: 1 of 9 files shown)


aentinger:
Good Morning @pennam :wave: :coffee: 
Possibly I've missed this, but did you extend the unit tests to also cover the new `CloudUnsignedInt`/`unsigned int` property?

pennam:
Hi @aentinger, you are right, i did not add them. Let me do it :muscle:

pennam:
@aentinger should be better now, i've added tests for encoding/decoding `CloudTime` property so it will cover also the `CloudUnsignedInt` type.

I was a bit unsure, but like for the other basic types (`CloudInt`, `CloudBool`...) i did not add any test for operator overloads.

aentinger:
>@aentinger should be better now, i've added tests for encoding/decoding CloudTime property so it will cover also the CloudUnsignedInt type.

Yes, that's perfectly fine.

PR #284, opened 2021-10-27
CloudSchedule

pennam:
Opening to start integration testing with cloud side.

/cc @eclipse1985

TODO:
- ~~Add property tests~~
- ~~Use const instead of numbers~~
- ~~Check CloudSchedule isActive() method~~
- ~~Timezone -> see https://github.com/pennam/ArduinoIoTCloud/tree/cloudschedule_timezone~~

### DIFF
--- a/examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
+++ b/examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
@@ -1,0 +1,54 @@
+/*
+  This sketch demonstrates how to use cloud schedule data type.
+
+  This sketch is compatible with:
+   - MKR 1000
+   - MKR WIFI 1010
+   - MKR GSM 1400
+   - MKR NB 1500
+   - MKR WAN 1300/1310
+   - Nano 33 IoT
+   - ESP 8266
+*/
+
+#include "arduino_secrets.h"
+#include "thingProperties.h"
+
+#if defined(ESP32)
+static int const LED_BUILTIN = 2;
+#endif
+
+void setup() {
+  /* Initialize serial and wait up to 5 seconds for port to open */
+  Serial.begin(9600);
+  for(unsigned long const serialBeginTime = millis(); !Serial && (millis() - serialBeginTime > 5000); ) { }
+
+  /* Configure LED pin as an output */
+  pinMode(LED_BUILTIN, OUTPUT);
+
+  /* This function takes care of connecting your sketch variables to the ArduinoIoTCloud object */
+  initProperties();
+
+  /* Initialize Arduino IoT Cloud library */
+  ArduinoCloud.begin(ArduinoIoTPreferredConnection);
+
+  setDebugMessageLevel(DBG_INFO);
+  ArduinoCloud.printDebugInfo();
+
+  /* Configure a schedule for LED. This should be done with Arduino create Scheduler widget */
+  unsigned int startingFrom = 1635786000;                /* From 01/11/2021 17:00           */
+  unsigned int untilTo = startingFrom + ( DAYS * 28 );   /* To   29/11/2021 17:00           */
+  unsigned int executionPeriod = MINUTES * 6;            /* For         6 minutes           */
+  unsigned int scheduleConfiguration =  134217770;       /* On monday wednesday and friday  */
+
+  led = Schedule(startingFrom, untilTo, executionPeriod, scheduleConfiguration);
+}
+
+void loop() {
+  ArduinoCloud.update();
+
+  /* Activate LED when schedule is active */
+  digitalWrite(LED_BUILTIN, led.isActive());
+  
+}
+
--- a/examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
+++ b/examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
@@ -63,7 +63,7 @@ void setup() {
   */
 void setupOneShotSchedule() {
 
-  unsigned int startingFrom = Schedule::getTimeFromString("2021 Nov 01 17:00:00");
+  unsigned int startingFrom = TimeService::getTimeFromString("2021 Nov 01 17:00:00");
   unsigned int until = startingFrom + ( DAYS * 1 );
   unsigned int activePeriod = MINUTES * 5;
 
@@ -79,7 +79,7 @@ void setupOneShotSchedule() {
   */
 void setupMinuteSchedule() {
 
-  unsigned int startingFrom = Schedule::getTimeFromString("2021 Nov 01 17:00:00");
+  unsigned int startingFrom = TimeService::getTimeFromString("2021 Nov 01 17:00:00");
   unsigned int until = startingFrom + ( DAYS * 1 );
   unsigned int activePeriod = SECONDS * 15;
   unsigned int repetitionPeriod = 1;
@@ -96,8 +96,8 @@ void setupMinuteSchedule() {
  */
 void setupHourlySchedule() {
 
-  unsigned int startingFrom = Schedule::getTimeFromString("2021 Nov 01 17:00:00");
-  unsigned int until = Schedule::getTimeFromString("2021 Nov 15 13:00:00");
+  unsigned int startingFrom = TimeService::getTimeFromString("2021 Nov 01 17:00:00");
+  unsigned int until = TimeService::getTimeFromString("2021 Nov 15 13:00:00");
   unsigned int activePeriod = MINUTES * 20;
   unsigned int repetitionPeriod = 1;
 
@@ -113,8 +113,8 @@ void setupHourlySchedule() {
  */
 void setupDailySchedule() {
 
-  unsigned int startingFrom = Schedule::getTimeFromString("2021 Nov 01 17:00:00");
-  unsigned int until = Schedule::getTimeFromString("2021 Nov 15 13:00:00");
+  unsigned int startingFrom = TimeService::getTimeFromString("2021 Nov 01 17:00:00");
+  unsigned int until = TimeService::getTimeFromString("2021 Nov 15 13:00:00");
   unsigned int activePeriod = HOURS * 2;
   unsigned int repetitionPeriod = 1;
 
@@ -137,7 +137,8 @@ void setupDailySchedule() {
  * Saturday  -> Inactive
  */
 void setupWeeklySchedule() {
-  unsigned int startingFrom = Schedule::getTimeFromString("2021 Nov 01 17:00:00");
+
+  unsigned int startingFrom = TimeService::getTimeFromString("2021 Nov 01 17:00:00");
   unsigned int until = startingFrom + ( DAYS * 30 );
   unsigned int executionPeriod = MINUTES * 3;
 
@@ -162,8 +163,8 @@ void setupWeeklySchedule() {
  */
 void setupMonthlySchedule() {
 
-  unsigned int startingFrom = Schedule::getTimeFromString("2021 Nov 01 17:00:00");
-  unsigned int until = Schedule::getTimeFromString("2021 Nov 15 13:00:00");
+  unsigned int startingFrom = TimeService::getTimeFromString("2021 Nov 01 17:00:00");
+  unsigned int until = TimeService::getTimeFromString("2021 Nov 15 13:00:00");
   unsigned int activePeriod = DAYS * 1;
   unsigned int dayOfMonth = 3;
 
@@ -179,8 +180,8 @@ void setupMonthlySchedule() {
  */
 void setupYearlySchedule() {
 
-  unsigned int startingFrom = Schedule::getTimeFromString("2021 Nov 06 17:00:00");
-  unsigned int until = Schedule::getTimeFromString("2041 Nov 06 13:00:00");
+  unsigned int startingFrom = TimeService::getTimeFromString("2021 Nov 06 17:00:00");
+  unsigned int until = TimeService::getTimeFromString("2041 Nov 06 13:00:00");
   unsigned int activePeriod = DAYS * 2;
   unsigned int dayOfMonth = 6;
 
--- a/examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
+++ b/examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
@@ -1,7 +1,7 @@
 /*
-  This sketch demonstrates how to use cloud schedule data type.
+  This sketch demonstrates how to use the cloud schedule variable type.
 
-  This sketch is compatible with:
+  This sketch is compatible with the following boards:
    - MKR 1000
    - MKR WIFI 1010
    - MKR GSM 1400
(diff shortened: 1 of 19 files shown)


aentinger (changes requested):
Hi @pennam :wave: :coffee: Can I ask you to add a example sketch show-casing how to actually use that property? From looking at the source code I don't really get it :confused:

pennam:
@aentinger sure!

pennam:
Hi @aentinger, i've added a simple sketch just to show how the cloud schedule property can be used.
I must admit that configuring a schedule from a sketch is really ugly, but the configuration part should be handled through the Arduino Create scheduler widget.

To define a schedule 4 field needs to be provided:
`frm`: uint32_t unix timestamo of schedule starting date and time
`to `: uint32_t unix timestamp of schedule end date and time. 0 if schedule has no end.
`len`: uint32_t duration of scheduled event
`msk`: uint32_t bitfield schedule configuration value.

` | 2bits scheduleUnit | 4bits scheduleType | 26 bits schedule Mask |`

```
enum class ScheduleUnit : int {
  Seconds      = 0,
  Minutes      = 1,
  Hours        = 2,
  Days         = 3
};

enum class ScheduleType : int {
  OneShot      = 0,
  FixedDelta   = 1,
  Weekly       = 2,
  Monthly      = 3,
  Yearly       = 4
};

Fixed schedule mask
delta value between events in ScheduleUnit

Weekly schedule mask
0000001 → Sunday
0000010 → Monday
0000100 → Tuesday
0001000 → Wednesday
0010000 → Thursday
0100000 → Friday
1000000 → Saturday

Monthly schedule mask
Day of the month 1 -31

Yearly schedule mask
| Months since January | Day of the month |
|         0-11         |       1-31       |
|         8 bit        |      8 bit       |
```

pennam:
Regarding the msk value i think i can easily add some function to ease creation of scheduler configs. Not sure what to do with frm and to fields... :thinking:

pennam:
Hi @aentinger, I still need to fix UnitTests after the last changes, in the meanwhile do you mind having a look to the new example and function to ease Schedule configuration ?

https://github.com/arduino-libraries/ArduinoIoTCloud/pull/284/commits/32292b5ac46c50bd302add21360af671f9f1a0ae also add support for timezone changes from the cloud.

Thanks

pennam:
rebased on master and squashed some commits to cleanup history

src/utility/time/TimeService.cpp
```diff
@@ -91,6 +91,52 @@ unsigned long TimeService::getTime()
...
+    0 /* tm_mday  */,
+    0 /* tm_mon   */,
+    0 /* tm_year  */,
+    0 /* tm_wday  */,
+    0 /* tm_yday  */,
+    0 /* tm_isdst */
+  };
+  static const char month_names[] = "JanFebMarAprMayJunJulAugSepOctNovDec";
+
+  sscanf(timeString.c_str(), "%d %s %d %d:%d:%d", &year, s_month, &day, &hour, &minutes, &seconds);
```
polldo, line 125: this function is public but there are no checks on the input
for example, if a user executes the following statement, the program will crash
`  Schedule::getTimeFromString("2021 November 01 17:00:00");  `

is this something we care about ?
also, I think that the time format should be documented somewhere

src/property/types/CloudSchedule.h
```diff
@@ -0,0 +1,417 @@
...
+          if ( ( (std::max(now , frm) - std::min(now , frm)) % delta ) <= len ) {
+            return true;
+          }
+        }
+      }
+      return false;
+    }
+
+    static unsigned int createOneShotScheduleConfiguration() {
+      return 0;
```
polldo, line 132: what should this function do?

src/utility/time/TimeService.h
```diff
@@ -55,8 +59,13 @@ class TimeService
 #endif
 
   unsigned long getRemoteTime();
+
   static bool isTimeValid(unsigned long const time);
 
+private:
+  unsigned long _timezone_offset;
```
polldo, line 66: the offset can be negative

src/utility/time/TimeService.cpp
pennam: yes, definitely. I was really unsure about this function, andas you can see here https://github.com/arduino-libraries/ArduinoIoTCloud/pull/284/commits/d317f5631b91f1479089ef519bc6c55b15b64c71 i had tried to use also std::get_time function to do this. Do you think it would be better?

I didn't wont to add much complex checks mostly because i think most people would use the widget to configure the schedule, but for sure i will add some lined to document how to use the function in the example and in the TimeService class

src/property/types/CloudSchedule.h
pennam: for non repeating schedule the configurazion mask is 0

src/utility/time/TimeService.h
pennam: Nice catch! thanks

src/utility/time/TimeService.cpp
pennam: or maybe jut provide a wrapper around std::get_time provigind the time string and also the information on how is formatted..

polldo: >as you can see here d317f56 i had tried to use also std::get_time function to do this. Do you think it would be better?

yes, using a std function with a standard time format seems better to me 

IMO all other alternatives are also good as long as they are documented (also the one you already implemented)

src/property/types/CloudSchedule.h
polldo: oh right!

polldo (approved):

src/utility/time/TimeService.cpp
pennam: fixed with 3cf18b29bd81944820adb793468b2f1b8a8d55e5

src/utility/time/TimeService.h
pennam: fixed with cd9a6642f0904630e9d5d747e28439e06b313b20

src/utility/time/TimeService.cpp
```diff
@@ -129,20 +128,66 @@ unsigned long TimeService::getLocalTime()
...
+  };
+
+  char s_month[16];
+  int month, day, year, hour, min, sec;
+  static const char month_names[] = "JanFebMarAprMayJunJulAugSepOctNovDec";
+  static const int input_length = 20;
+  static const int input_parameters = 6;
+  unsigned long time = 0;
+
+  if(input != nullptr && input.length() == input_length) {
```
polldo, line 153:
```C++
if (input == nullptr || input.length() != input_length) {
       DEBUG_ERROR("ArduinoIoTCloudTCP::%s invalid input length", __FUNCTION__);
       return 0
}
```
I think it'd be more readable this way

src/utility/time/TimeService.cpp
```diff
@@ -129,20 +128,66 @@ unsigned long TimeService::getLocalTime()
...
+
+    int result = sscanf(input.c_str(), "%d %s %d %d:%d:%d", &year, s_month, &day, &hour, &min, &sec);
+
+    if(result != input_parameters)
+    {
+      DEBUG_ERROR("ArduinoIoTCloudTCP::%s invalid input parameters number", __FUNCTION__);
+      return 0;
+    }
+
+    if(strlen(s_month) > 3) {
```
polldo, line 163: we may want to check that `strlen(s_month)` is exactly equal to 3 ?

src/utility/time/TimeService.cpp
```diff
@@ -129,20 +128,66 @@ unsigned long TimeService::getLocalTime()
...
+      DEBUG_ERROR("ArduinoIoTCloudTCP::%s invalid month name length, use %s", __FUNCTION__, month_names);
+      return 0;
+    }
+
+    month = (strstr(month_names, s_month) - month_names) / 3;
```
polldo, line 168: nice pointer arithmetic, but there's a problem if the month is not contained in `month_names` 
I would add a `!= nullptr` check on the result of strstr before using it

examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
```diff
@@ -0,0 +1,229 @@
...
+  * Until         2021 11 02 17:00:00
+  */
+void setupOneShotSchedule() {
+
+  unsigned int startingFrom = TimeService::getTimeFromString("2021 Nov 01 17:00:00");
+  unsigned int until = startingFrom + ( DAYS * 1 );
+  unsigned int activePeriod = MINUTES * 5;
+
+  /* Warning: there is no cross check between until and activePeriod */
+  unsigned int scheduleConfiguration =  Schedule::createOneShotScheduleConfiguration();
```
aentinger, line 71: I was thinking: Why not create a `typedef` to hide the fact that a schedule is nothing more than an `unsigned int`? If at some point in time it becomes a class there's no need change anything, as it already appears like a class due to the `typedef`, i.e.
```C++
typedef unsigned int CloudScheduleType;
```
This could be combined with a `static_assert` to ensure that the cloud type always has the minimum required value, i.e.
```C++
static_assert(sizeof(CloudScheduleType int) >= 4", Error, sizeof(CloudScheduleType) < 4, can't use CloudSchedule on this platform."), 
```

src/ArduinoIoTCloud.h
```diff
@@ -145,7 +147,9 @@ class ArduinoIoTCloudClass
 
     ConnectionHandler * _connection = nullptr;
     PropertyContainer _property_container;
-    TimeService _time_service;
+    TimeService * _time_service = ArduinoIoTCloudTimeService();
```
aentinger, line 150: Is there any specific reason why this is now a pointer? It looks like you want to employ the `Singleton` design pattern. Why? Usually (except for very few use cases) `Singleton`s are to be avoided.

Also please do all initialisation within the constructor, or all initialisation within the class body, but not mixing between those two options.

extras/test/src/util/PropertyTestUtil.cpp
```diff
@@ -16,3 +18,13 @@ unsigned long getTime()
 {
   return 0;
 }
+
+unsigned long TimeService::getLocalTime()
+{
+  return getTime();
+}
+
+TimeService* ArduinoIoTCloudTimeService() {
```
aentinger, line 27: This looks somewhat like you want to use a `Singleton` (also see comment below). In general I'd advise to avoid singletons but if you must have them then please follow this pattern:
```C++
/* Singleton.h */
class Singleton
{
public:
  Singleton(Singleton &) = delete;
  void operator = (Singleton &) = delete;

  static Singleton & instance();
  static void destroy();
private:
  static Singleton * _p_instance;
};
/* Singleton.cpp */
Singleton * Singleton::_p_instance{nullptr};
Singleton & Singleton::instance()
{
  if (!_p_instance) {
    _p_instance = new Singleton();
  }
  return *_p_instance;
}

void Singleton::destroy()
{
  delete _p_instance;
  _p_instance = nullptr;
}
```

aentinger (commented):
I love how this PR comes together :heart_eyes:, however, since I'm a nitpicker I still found a couple of suggestions for you :wink:

src/utility/time/TimeService.cpp
pennam: Thanks @polldo, please see 6b9602520d013a1f670abf83e9f08ffe17bb22de

src/utility/time/TimeService.cpp
pennam: oh yes -> 6b9602520d013a1f670abf83e9f08ffe17bb22de

src/utility/time/TimeService.cpp
pennam: Yes! all credits go to @aentinger :wink: . I've only re-used it badly :smiling_face_with_tear: I've never had to use pointer aritmetic before and i've discovered that null pointer aritmetic is quite an hot topic!! ->https://github.com/arduino-libraries/ArduinoIoTCloud/commit/6b9602520d013a1f670abf83e9f08ffe17bb22de

pennam:
>  since I'm a nitpicker I still found a couple of suggestions for you wink

That's the reason why i continue bothering you with my review requests! Thanks!

examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
pennam: I'm not sure i get the point, `Schedule` and `CloudSchedule` are already two classes; in the example `scheduleConfiguration` is just one parameter needed to initialize the schedule. That said i think using a typedef for the scheduleConfiguration and the other input parameters is really a good idea so i've created a commit you can check. https://github.com/pennam/ArduinoIoTCloud/commit/2b63f01f2e41c2ad9d75729087eb95bbbf0960e9

src/utility/time/TimeService.cpp
aentinger: Yeah well, I totally copied that from somewhere else :blush:

examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
aentinger: No, you've gotten me perfectly. :+1: for `ScheduleTimeType` and `ScheduleConfigurationType`.

src/ArduinoIoTCloud.h
aentinger: I see. Still one small improvement would be to operate with references `&` instead of pointers `*`, then you never need to worry that a pointer can be zero, i.e.
```C++
TimeService & ArduinoIoTCloudTimeService() {
  static TimeService _timeService_instance;
  return _timeService_instance;
}
```

aentinger: Related: Why do you have `ArduinoIoTCloudTimeService` twice in code (Once in production code, once in test code). Would it not be sufficient to have it defined only once? I do not see MCU-specific code in that statement that would not compile for host.

examples/ArduinoIoTCloud-Schedule/ArduinoIoTCloud-Schedule.ino
pennam: -> see https://github.com/arduino-libraries/ArduinoIoTCloud/pull/284/commits/210edd275792da456221fb8d413a884cbcb624be

src/ArduinoIoTCloud.h
pennam: No MCU-specific code, but adding TimeService.cpp to the test system will require some extra work because then we will need also to stub/fake a bit on NTPUtils and maybe also RTC. So at the end i've decided to duplicate the function to make the build system happy and run the tests.

The only property that is actually using TimeService is CloudSchedule so i don't think adding the whole class would be really beneficial for the test coverage since CloudSchedule is implementing its own getTime() fake function in test_CluodSchedule.cpp file.

pennam: > operate with references & instead of pointers *

please see https://github.com/arduino-libraries/ArduinoIoTCloud/pull/284/commits/920e920a7f4125e6272b8c6ba814301eb9f0442f

extras/test/src/util/PropertyTestUtil.cpp
pennam: implemented with the pattern above using references -> see https://github.com/arduino-libraries/ArduinoIoTCloud/commit/920e920a7f4125e6272b8c6ba814301eb9f0442f

aentinger:
I still suggest to [initialise member variables](https://github.com/arduino-libraries/ArduinoIoTCloud/pull/284/files#diff-2a08667a46c0a45cd028e94148001c11fd307f4d237de0a2f6166c8873777362R151-R152) within the constructor (to stay consistent with the current pattern), but apart from this, this is done imho.

aentinger (approved):
Love it :+1:

pennam:
we need to upgrade that ESP core! :smiling_face_with_tear:

pennam:
rebased and squashed some commits to cleanup history before merging.

PR #289, opened 2021-11-11
Fix OTA properties synchronization

pennam:
Switching to  ArduinoIoTCloud 1.3.0 from earlier version can lead to an OTA loop caused by bad synchronization of OTA_REQ property. The OTA loop occurs when the resulting CBOR message created calling sendPropertiesToCloud() is greater than 256 byte and thus it gets splitted in multiple parts. If the OTA variables are not synched within the firs part of the message the OTA process restart before the remainig properties are synched.

With this PR the OTA properties are synched all together within the same CBOR message ensuring a correct synchronization.

Updating an application builded with an ArduinoIoTCloud library < 1.3.0 with an OTA will still lead to an OTA process executed two times.

Thanks @eclipse1985 for spotting this.

/cc @zmoog

### DIFF
--- a/src/ArduinoIoTCloudTCP.cpp
+++ b/src/ArduinoIoTCloudTCP.cpp
@@ -34,7 +34,6 @@
 #endif
 
 #if defined(ARDUINO_PORTENTA_H7_M7) || defined(ARDUINO_PORTENTA_H7_M4)
-#  include <algorithm>
 #  include "tls/utility/SHA256.h"
 #  include <stm32h7xx_hal_rtc_ex.h>
 #  include <WiFi.h>
@@ -42,7 +41,7 @@
 
 #include "utility/ota/OTA.h"
 #include "utility/ota/FlashSHA256.h"
-
+#include <algorithm>
 #include "cbor/CBOREncoder.h"
 
 #include "utility/watchdog/Watchdog.h"
@@ -516,7 +515,7 @@ ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_Connected()
         /* Clear the request flag. */
         _ota_req = false;
         /* Transmit the cleared error and request flags to the cloud. */
-        sendPropertiesToCloud();
+        sendOTAPropertiesToCloud();
         /* Call member function to handle OTA request. */
         onOTARequest();
       }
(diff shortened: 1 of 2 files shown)


src/ArduinoIoTCloudTCP.cpp
```diff
@@ -581,6 +580,30 @@ void ArduinoIoTCloudTCP::sendPropertiesToCloud()
...
+{
+  sendPropertyContainerToCloud(_property_container);
+}
+
+void ArduinoIoTCloudTCP::sendOTAPropertiesToCloud()
+{
+  PropertyContainer to_property_container;
+  PropertyContainer from_property_container = _property_container;
+
+  std::list<String> ota_property_list {"OTA_CAP", "OTA_ERROR", "OTA_SHA256", "OTA_URL", "OTA_REQ"};
```
aentinger, line 593: Minor nitpick: `ota_property_list` can be const, which does entail using `cbegin()`, `cend()` in `std::for_each` below, i.e.
```C++
std::list<String> const ota_property_list {"OTA_CAP", "OTA_ERROR", "OTA_SHA256", "OTA_URL", "OTA_REQ"};
  std::for_each(ota_property_list.cbegin(),
                ota_property_list.cend(),
/* ... */
```

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -581,6 +580,30 @@ void ArduinoIoTCloudTCP::sendPropertiesToCloud()
...
 }
 
+void ArduinoIoTCloudTCP::sendPropertiesToCloud()
+{
+  sendPropertyContainerToCloud(_property_container);
+}
+
+void ArduinoIoTCloudTCP::sendOTAPropertiesToCloud()
+{
+  PropertyContainer to_property_container;
```
aentinger, line 590: Those 2 names (`to_property_container`, `from_property_container`) got me confused on the first time. How about naming them like:
* `to_property_container` -> `ota_property_container`
* `from_property_container` -> `full_property_container`

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -581,6 +580,30 @@ void ArduinoIoTCloudTCP::sendPropertiesToCloud()
...
+  PropertyContainer from_property_container = _property_container;
```
aentinger, line 591: Also: Is it really necessary to copy the full property container here? Since you do not change it I don't see why you shouldn't be able to use `_property_container` in the `std::for_each` loop below.

aentinger (approved):
LGTM :+1: Minor nitpicks - use the feedback at your own caution - overall looks good :+1:

src/ArduinoIoTCloudTCP.cpp
pennam: That was also my intention, but doing so compiler is complaining...
```
/home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloudTCP.cpp: In member function 'void ArduinoIoTCloudTCP::sendOTAPropertiesToCloud()':
/home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloudTCP.cpp:596:19: error: capture of non-variable 'ArduinoIoTCloudClass::_property_container' 
                 [&_property_container, &to_property_container ] (String const & name)
                   ^~~~~~~~~~~~~~~~~~~
In file included from /home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloudTCP.h:27:0,
                 from /home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloudTCP.cpp:25:
/home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloud.h:147:23: note: 'PropertyContainer ArduinoIoTCloudClass::_property_container' declared here
     PropertyContainer _property_container;
                       ^~~~~~~~~~~~~~~~~~~
/home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloudTCP.cpp: In lambda function:
/home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloudTCP.cpp:598:45: error: 'this' was not captured for this lambda function
                   Property* p = getProperty(_property_container, name);
                                             ^~~~~~~~~~~~~~~~~~~
/home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloudTCP.cpp:598:45: error: invalid use of non-static data member 'ArduinoIoTCloudClass::_property_container'
In file included from /home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloudTCP.h:27:0,
                 from /home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloudTCP.cpp:25:
/home/pennam/Arduino/libraries/ArduinoIoTCloud/src/ArduinoIoTCloud.h:147:23: note: declared here
     PropertyContainer _property_container;
                       ^~~~~~~~~~~~~~~~~~~
```
I think the point is that `_property_container` is not a local variable, but this was actually my firs lambda expression so i'm not really sure. :sweat_smile:

aentinger: That one is easy :+1: you need to capture `this` within the lamda :wink: 
```C++
[this, &to_property_container ] (String const & name)
```

pennam: nice, I could have spent weeks trying to find this solution. Thanks Alex.  :rocket: 

Now i'm ready to fix it, merge everything and release a new library version.

PR #292, opened 2021-12-09
Minor change: Schedule example sketch: Fix supported boards list

pennam:
removing MKR WAN boards from the header of ArduinoIoTCloud-Schedule.ino example sketch because MKR WAN have no information about time so the scheduler wont work.


aentinger:
I've just taken a look and I think maybe use a "smaller" PR title, I was looking at my email and totally expected to see the removal of MKR WAN support from the AIOTC stack, but you are effectively removing a comment. Maybe "Minor Fix: Cleaning up documentation re MKR WAN/CloudSchedule" or smth like that.

pennam:
:cold_sweat: sure didn't though about it. I will reword also the commit message!

aentinger:
Thank you :bow: :wink:

PR #293, opened 2021-12-17
Avoid multivalue properties split

pennam:
Problem: 
with the current logic of `CBOR` message creation is actually possible that a multivalue property is splitted in two consecutive messages. This would break any multivalue property widget in create dashboard because is expected that all the values of the same multivalue property are updated at the same time.

Proposed solution:
Adding `CHECK_CBOR_MULTI` macro is possible to throw a `CborErrorTooManyItems` error when the whole multivalue property does not fit into the `CBOR` message. As a consequence the `CBOR` message is discarded (not sended to the cloud) and all the properties included in the discarded message and markes as `appended_but_not_sended` are scheduled for re-send. Furthermore the `encoded_properties_message_limit` is updated in order to send the correct number of properties to avoid re-trigger the error. The excluded property will be then sended in the next message. The same logic applies when the error occurs closing the `CBOR` container.

First run:
```
prop_a
prop_b
prop_c
multi_prop_d.1
multi_pror_d.2 <- CborErrorTooManyItems
```
Second run:
```
prop_a
prop_b
prop_c
```
Third run:
```
multi_prop_d.1
multi_pror_d.2 
...
```

Other changes:
Commits https://github.com/arduino-libraries/ArduinoIoTCloud/commit/eb41da82c0e31a7c8908216a2a695d985de93250  and https://github.com/arduino-libraries/ArduinoIoTCloud/commit/ab527a150f53dea7a408887f239a446db3ffadbe are addressing a different but in some way correlated issue. If the propety container has a number of high change rate properties in the first positions and this properties are filling the `CBOR` message buffer these high change rate properties are updated each 
`sendPropertiesToCloud` call leaving out the other properties.

This can be easily reproduced creating a sketch with 18 float properties and a loop like the one below. `atest_12` is never updated.

```
void loop() {
  ArduinoCloud.update();
 
      atest_0++;
      atest_1++;
      atest_2++;
      atest_3++;
      atest_4++;
      atest_5++;
      atest_6++;
      atest_7++;
      atest_8++;
      atest_9++;
      atest_A++;
      atest_B++;
      atest_C++;
      atest_D++;
      atest_E++;
      atest_F++;
      atest_10++;
      atest_11++;
      atest_12++;
      delay(500);
}
```
Lastly since i've added the `appended_but_not_sended` i've also included commit https://github.com/arduino-libraries/ArduinoIoTCloud/commit/14eb4799320eacd0ca9d967aa6a988935f97c8d4 that does not change any behaviour of the library, but it makes a bit clear that we need to echo back the received properties to the cloud even if unchanged.

/cc @eclipse1985


src/cbor/CBOREncoder.cpp
```diff
@@ -45,24 +46,68 @@ CborError CBOREncoder::encode(PropertyContainer & property_container, uint8_t *
...
+  if(current_property_index >= property_container.size())
+    current_property_index = 0;
+
+  PropertyContainer::iterator iter = property_container.begin();
+  std::advance(iter, current_property_index);
+
+  for(; iter != property_container.end(); iter++)
+  {
+    Property * p = * iter;
+    bool maximum_number_of_properties_reached = (num_encoded_properties >= encoded_properties_message_limit) && (encoded_properties_message_limit != -1);
```
aentinger, line 61:
```suggestion
    bool const maximum_number_of_properties_reached = (num_encoded_properties >= encoded_properties_message_limit) && (encoded_properties_message_limit != -1);
```
Use [const-correctness](https://isocpp.org/wiki/faq/const-correctness#overview-const) as much as possible :wink:

src/cbor/CBOREncoder.cpp
```diff
@@ -45,24 +46,68 @@ CborError CBOREncoder::encode(PropertyContainer & property_container, uint8_t *
...
+    bool cbor_encoder_error = (error != CborNoError);
```
aentinger, line 62:
```suggestion
    bool const cbor_encoder_error = (error != CborNoError);
```
Same as above. Btw, I love it that you are "encoding" the meaning of a boolean operation in the name of the boolean variable. Makes the if statement below so much more easily readable.

src/cbor/CBOREncoder.cpp
```diff
@@ -45,24 +46,68 @@ CborError CBOREncoder::encode(PropertyContainer & property_container, uint8_t *
...
+
+  if(current_property_index >= property_container.size())
+    current_property_index = 0;
+
+  PropertyContainer::iterator iter = property_container.begin();
+  std::advance(iter, current_property_index);
+
+  for(; iter != property_container.end(); iter++)
+  {
+    Property * p = * iter;
```
aentinger, line 60: I'm trying to make up my mind whether or not this "duplication" of `iter` is really necessary. Possibly if you rename `iter` to `prop_iter` and then access via `(*prop_iter)->`. Granted, this syntax (`(*exp)->`) is somewhat ugly. I leave it for you to decide :shrug:

src/cbor/CBOREncoder.cpp
```diff
@@ -45,24 +46,68 @@ CborError CBOREncoder::encode(PropertyContainer & property_container, uint8_t *
...
+    /* Trim the number of properties to be included in the next message to avoid multivalue property split */
+    encoded_properties_message_limit = num_encoded_properties;
     return error;
+  }
+
+  error = cbor_encoder_close_container(&encoder, &arrayEncoder);
+  if (CborNoError != error)
+  {
+    /* Trim the number of properties to be included in the next message to avoid error closing container */
+    encoded_properties_message_limit = num_encoded_properties - 1;
```
aentinger, line 87: The change feels like it uses `encoded_properties_message_limit` as both counter AND status variable. Can you think of a way to change the PR to `encoded_properties_message_limit` only as a counter (and possibly rename it) and use a boolean variable to cover the cases `CBOR_ENCODER_NO_PROPERTIES_LIMIT` and `-1`. If `num_encoded_properties - 1` is also a valid state than this should be encoded as a [FSM](https://en.wikipedia.org/wiki/Finite-state_machine) making the states and state transitions explicit. Right now states and state transitionings are hidden in the business logic and really hard to see.

aentinger (commented):
I think my most important feedback it to try and redesign the encoding logic into a FSM with named states, so that it's very clear in which state you are at all times and also the transitioning paths between the states are clearly visible. For a sample FSM pattern see [here](https://github.com/arduino-libraries/ArduinoIoTCloud/blob/master/src/ArduinoIoTCloudTCP.cpp#L318-L328). :bow: As always it's up to you to act on my feedback, please see it as a recommendation, not a request.

pennam:
> I think my most important feedback it to try and redesign the encoding logic into a FSM with named states, so that it's very clear in which state you are at all times and also the transitioning paths between the states are clearly visible. For a sample FSM pattern see [here](https://github.com/arduino-libraries/ArduinoIoTCloud/blob/master/src/ArduinoIoTCloudTCP.cpp#L318-L328). bow As always it's up to you to act on my feedback, please see it as a recommendation, not a request.

got it! i will take some time to think about it, but i agree that splitting the encoding process in basic operation would make the code more readable and the logic more explicit!. Thanks.

aentinger:
Great :+1: There's one thing that came to my mind additionally, which is that I'd suggest to capture the desired "output" within a unit test, i.e. you replicate your scenario where a multi-value property would be cut off, but isn't due to this change. Then you've got a much easier time refactoring this code, since you can run the tests after every change and be assured, that everything is still working as it should :wink:

src/cbor/CBOREncoder.cpp
pennam: Fixed. https://github.com/arduino-libraries/ArduinoIoTCloud/pull/293/commits/eedffb304d4fd5a1890e7f030080a5f23d49cad3

src/cbor/CBOREncoder.cpp
pennam: Same as above also fo me.

src/cbor/CBOREncoder.cpp
pennam: I personally i don't like either of them, but i like a little bit more `Property * p = * iter;` despite the variable duplication.

pennam:
> Great +1 There's one thing that came to my mind additionally, which is that I'd suggest to capture the desired "output" within a unit test, i.e. you replicate your scenario where a multi-value property would be cut off, but isn't due to this change. Then you've got a much easier time refactoring this code, since you can run the tests after every change and be assured, that everything is still working as it should wink

Added a test here https://github.com/arduino-libraries/ArduinoIoTCloud/pull/293/commits/c620ad12c5e29bb8ca2ecefe38ff13556b9f204e

pennam:
Hi @aentinger did you had a chance to look at this commit https://github.com/arduino-libraries/ArduinoIoTCloud/pull/293/commits/eedffb304d4fd5a1890e7f030080a5f23d49cad3

aentinger:
Good Morning @pennam ☕ :wave:
I did look at the commit you referenced above and also at the complete PR. I feel a bit guilty like I intentionally misled you with my feedback, because now I'm going to ask you to cut a bit back on the states within your state machine. Basically the only states that need to be distinct states are the ones that can fail during a single execution of `CborEncoder::encode`.

An example: Neither `handle_InitPropertyEncoder()` nor `handle_OpenCBORContainer()` can fail, consequently they could be grouped in one state(`EncoderState::Prepare`)/function `handle_Prepare()`, i.e.
```C++
void handle_Prepare()
{
  handle_InitPropertyEncoder(propertyEncoder);
  handle_OpenCBORContainer(propertyEncoder, data, size);
  /* ... */
}
```
Otherwise you are pretty much there.

src/ArduinoIoTCloudLPWAN.cpp
```diff
@@ -149,8 +149,9 @@ void ArduinoIoTCloudLPWAN::sendPropertiesToCloud()
 {
   int bytes_encoded = 0;
   uint8_t data[CBOR_LORA_MSG_MAX_SIZE];
+  static unsigned int last_checked_property_index = 0;
```
aentinger, line 152: This now is where you save the state for the `CBOREncoder::encode` state machine. I'd suggest to rather store it as a protected (not private, because both `ArduinoIoTCloudLPWAN` and `ArduinoIoTCloudTCP` need to access it) **member variable** of class `ArduinoIoTCloud`. Otherwise you've just created a hidden state variable. In my experience, everything that controls state in a C++ class should be listed as a member in the class body.

library.properties
```diff
@@ -1,5 +1,5 @@
 name=ArduinoIoTCloud
-version=1.3.1
+version=1.4.0
```
aentinger, line 2: Do you bump the library version number in this PR? It's of course possible, but the/my ususal flow is to merge the PR, than make a dedicated "release" commit.

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -596,12 +596,14 @@ void ArduinoIoTCloudTCP::sendPropertyContainerToCloud(PropertyContainer & proper
 
 void ArduinoIoTCloudTCP::sendPropertiesToCloud()
 {
-  sendPropertyContainerToCloud(_property_container);
+  static unsigned int last_checked_property_index = 0;
```
aentinger, line 599: Again, this is hidden state. Please consider moving it to the class body.

src/cbor/CBOREncoder.cpp
aentinger: Works for me 👍 😉

aentinger (changes requested):
We are getting there, only cosmetic changes are left 😉

library.properties
pennam: ups, this was due to a wrong rebase, 1.4.0 is already released

src/ArduinoIoTCloudLPWAN.cpp
pennam: now i got it! plese see https://github.com/arduino-libraries/ArduinoIoTCloud/pull/293/commits/5576a608dba2186e87c8cdc8f8f81be42a0687fd

pennam:
> An example: Neither `handle_InitPropertyEncoder()` nor `handle_OpenCBORContainer()` can fail, consequently they could be grouped in one state(`EncoderState::Prepare`)/function `handle_Prepare()`, i.e.

I did not group this two states because even if they cannot fail they are doing different things. `InitPropertyEncoder` is called only at the beginning of each property chunck and is resetting `encoded_property_limit` counter and `property_limit_active` flag. `OpenCBORContainer` is called after `InitPropertyEncoder`, but also after `TrimAppend` and `TrimClose` and in this case we don't want to reset the conter and the flag.

I've grouped `FinishAppend` and `AdvancePropertyContainer` instead. See https://github.com/arduino-libraries/ArduinoIoTCloud/pull/293/commits/78b31586b9117d5483704902ea38543220f10ad1

aentinger:
Makes sense :+1: Thank you very much for acting on my feedback :bow:.

aentinger (approved):

pennam:
> Makes sense +1 Thank you very much for acting on my feedback bow.

Thanks for your help @aentinger :muscle:

PR #297, opened 2022-01-12
Thing id discovery protocol

pennam:
The Goal of this PR is to remove the hardcoded thing_id from the sketch and improve OTA flow.

Properties spit:
 - Propetries are splitted in device_properties and thing_properties with 2 separate containers.
 - Thing configuration and OTA will use device_properties.
 - User data will use thing_properties.

Thing_id: 
 - No hardcoded thing_id, the board will receive the thing_id from the cloud after subscribing to device_id topic.
 - After receiving the thing_id the general flow is unchanged and the board will start listen on thing_shadow_topic for LastValues and on thing_input_topic for InputValues.

OTA flow:
 - Send OTA_CAP, OTA_ERROR, and OTA_SHA256 as soon as connected to the broker to close any pending OTA process.
 - Use a dedicated topic for OTA and thing configuration.

https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297/commits/f2d0bdf315366e9d0f97ae9844d3c76cdf9662c4 and https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297/commits/93057615f25944728c557c2ece32ba89ea4c03ef are here only for testing purpose and will be dropped before merging.

Thing ID configuration :heavy_check_mark: 
Attach/Detach action 
OTA :heavy_check_mark: 
Deferred OTA :heavy_check_mark:


src/ArduinoIoTCloudTCP.cpp
```diff
@@ -763,14 +763,16 @@ void ArduinoIoTCloudTCP::sendPropertiesToCloud()
 }
 
 #if OTA_ENABLED
-void ArduinoIoTCloudTCP::sendOTAPropertiesToCloud()
+void ArduinoIoTCloudTCP::sendOTAPropertiesToCloud(bool include_ota_req)
```
aentinger:
```diff
-bool include_ota_req
+bool const include_ota_req
```
Minor suggestions, basically just use [const-correctness](https://isocpp.org/wiki/faq/const-correctness) as much as possible.

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -763,14 +763,16 @@ void ArduinoIoTCloudTCP::sendPropertiesToCloud()
...
 {
   PropertyContainer ota_property_container;
   unsigned int last_ota_property_index = 0;
 
-  std::list<String> const ota_property_list {"OTA_CAP", "OTA_ERROR", "OTA_SHA256"};
-  std::for_each(ota_property_list.cbegin(),
-                ota_property_list.cend(),
+  std::list<String> ota_property_list {"OTA_CAP", "OTA_ERROR", "OTA_SHA256"};
+  if (include_ota_req)
```
aentinger: I'd suggest to either enclose the expression below within braces, i.e.
```C++
if (include_ota_req) {
  ota_property_list.push_back("OTA_REQ");
}
```
or to leave one free line below the statement before and the statement after. Otherwise one get's all expressions clustered up and its not easy to see what belongs to the if statement and what not.

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -69,6 +69,11 @@ extern "C" void updateTimezoneInfo()
   ArduinoCloud.updateInternalTimezoneInfo();
 }
 
+extern "C" void setThingIdOutdated()
```
aentinger, line 72: Where is this function called? Why is there an `extern "C"` declaration necessary? (Looks like it's passed to C code as a callback function).

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -497,16 +501,13 @@ ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_CheckDeviceConfig()
...
 
-  if(getThingIdOutdatedFlag())
+  if(_deviceSubscribedToThing == true)
   {
-    if(_deviceSubscribedToThing == true)
-    {
-      /* Unsubscribe from old things topics and go oi with a new subsctiption */
-      _mqttClient.unsubscribe(_shadowTopicIn);
-      _mqttClient.unsubscribe(_dataTopicIn);
+    /* Unsubscribe from old things topics and go oi with a new subsctiption */
```
aentinger: Typo in comment, `oi` -> `on` (I guess).

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -585,24 +726,33 @@ void ArduinoIoTCloudTCP::handleMessage(int length)
...
     DEBUG_VERBOSE("ArduinoIoTCloudTCP::%s [%d] last values received", __FUNCTION__, millis());
-    CBORDecoder::decode(_property_container, (uint8_t*)bytes, length, true);
-    sendPropertiesToCloud();
+    CBORDecoder::decode(_thing_property_container, (uint8_t*)bytes, length, true);
     _time_service.setTimeZoneData(_tz_offset, _tz_dst_until);
     execCloudEventCallback(ArduinoIoTCloudEvent::SYNC);
     _last_sync_request_cnt = 0;
     _last_sync_request_tick = 0;
-    _state = State::Connected;
+    _next_state = State::Connected;
```
aentinger, line 751: Is it really necessary to hold `_next_state` as a private member variable and therefore make it part of the overall state of an `ArduinoIoTCloudTCP` object? I'm sceptical of any additional member variable since each member variable equals at least two (bool) or many more possible states. Therefore many member variables equal exploding state space which makes it more easy to introduce new bugs and more hard to test all possible state transition paths.

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -661,18 +661,18 @@ ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_Connected()
         /* Clear the request flag. */
         _ota_req = false;
         /* Transmit the cleared request flags to the cloud. */
-        sendClearedOTARequestToCloud();
+        sendDevicePropertyToCloud("OTA_REQ");
```
aentinger, line 677: I love this commit. Reducing code size while increasing readability. Top work! :bow:

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -256,8 +256,8 @@ int ArduinoIoTCloudTCP::begin(bool const enable_watchdog, String brokerAddress,
   addPropertyReal(_ota_cap, _device_property_container, "OTA_CAP", Permission::Read);
   addPropertyReal(_ota_error, _device_property_container, "OTA_ERROR", Permission::Read);
   addPropertyReal(_ota_img_sha256, _device_property_container, "OTA_SHA256", Permission::Read);
-  addPropertyReal(_ota_url, _device_property_container, "OTA_URL", Permission::ReadWrite).onSync(CLOUD_WINS);
-  addPropertyReal(_ota_req, _device_property_container, "OTA_REQ", Permission::ReadWrite).onSync(CLOUD_WINS);
+  addPropertyReal(_ota_url, _device_property_container, "OTA_URL", Permission::ReadWrite);
```
aentinger, line 260: This is removed why? Because now the updating of the values are enforced by discrete functions during the normal program flow?

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -426,9 +425,7 @@ ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_SendDeviceProperties()
     return State::Disconnect;
   }
 
-#if OTA_ENABLED
```
aentinger: Will this be a problem for devices which do not support OTA, i.e. MKR 1000, MKR GSM 1400, ... ?

aentinger (commented):
This PR is quite hard for me to review due to lack of knowledge of the background info re thing id update, consequently the value of my advice is probably limited. Nonetheless, here are a couple of thoughts :bow:.

src/ArduinoIoTCloudTCP.cpp
pennam: Yes you are right, this was added to support deferred ota some tima ago. But with this new flow is noy needed anymore.
To keep OTA_URL and OTA_REQ in sync with the cloud we proceed like this:
1 - The device sends to the cloud  OTA_CAP, OTA_ERROR and OTA_SHA
2 - The device subscribes to the device topic
3 - Cloud replies with THING_ID OTA_REQ and OTA_URL

src/ArduinoIoTCloudTCP.cpp
pennam: Thanks :muscle:

src/ArduinoIoTCloudTCP.cpp
pennam: This function is used as callback function fot the update event of the `_thing_id` property. Each time a new thing_id value is received this function is called to restart a new configuration process. It should not be necessary, i think i've copy pasted from 

`extern "C" void updateTimezoneInfo()`

that i have again copy pasted from

`extern "C" unsigned long getTime()`

but i think it is safe to remove all three `extern "C"` declarations

src/ArduinoIoTCloudTCP.cpp
pennam: i'll stick a post-it on my monitor so i will not forget it anymore :sweat_smile:

pennam: i'm lucky enough this change was outdated and not included in the final changes

src/ArduinoIoTCloudTCP.cpp
pennam: I think that also this line is outdated and not part of the final changes, but this made me doublecheck the code because after your question i was a bit unsure about it, so thanks.

In the final changes there are two versions of the `sendDevicePropertiesToCloud()` the first one sends all the device properties and it is availabe for all boards, the second one accept a string as input to send one property at time ad is used and availabe only for boards supporting OTA.

src/ArduinoIoTCloudTCP.cpp
aentinger: Allright, but where is this function registered as a callback?

src/ArduinoIoTCloudTCP.cpp
pennam: fully agree, this was part of the initial changes i've done to support OTA with this new flow, then i've changed this function in this commit https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297/commits/e6b252c280d27929e9d0eeb1995d4cf29dd624e1 and again here https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297/commits/98ed735b7136c7acbc22cd633424848be62da1d4

src/ArduinoIoTCloudTCP.cpp
pennam: yes, it was already fixed here https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297/commits/defe4dd583f8fa16ccd286f7ff63b4e732ec670e i know it has nothig to do with delay macros... i will create a separate commit and squash it to the original one

src/ArduinoIoTCloudTCP.cpp
pennam: need to doublecheck, but i'm quite confident this commit can be dropped, because `State::CheckDeviceConfig` logic as changed after this commit.

src/ArduinoIoTCloudTCP.cpp
pennam: Here https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297/commits/db3e9433c612998ddf0a0c3e9f3027a446d9c588 i think i've messed up a bit rebasing this firsts commits. I'll fix it keeping 

`  addPropertyReal(_thing_id, _device_property_container, "thing_id", Permission::ReadWrite)` in the first and adding

`onUpdate(setThingIdOutdated);` in the second one

pennam: I've also added a new commit to remove the extern `"C" declaration` see https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297/commits/0541f7bbd19214c4a1402d20631786ff4b8e722c

src/ArduinoIoTCloudTCP.cpp
pennam: Removed the fix from here https://github.com/arduino-libraries/ArduinoIoTCloud/commit/defe4dd583f8fa16ccd286f7ff63b4e732ec670e and squashed directly here https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297/commits/a0916cdce61fd0ab264358dbff68328eea8eeb56

src/ArduinoIoTCloudTCP.cpp
pennam: i've dropped this commit since at the end of the whole work it wasn't necessary.

pennam:
i've found a stray debug print, in the first commit so i will force push again in a while

pennam:
dropping https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297/commits/912b06868d8e4f144fa4477de4f805ea99996c65 and force-pushing again since setThingId() it is still needed for LORA device

PR #298, opened 2022-01-13
Add retry delay on topic subscription failures

pennam:
Most of the time subscripion failures are caused by wrong device configuration, so retry to subscribe immediately is not solving the problem. 

Adding a small delay (1 second )will make the debug error more readable by the user and it will also reduce the amout of subscription requests received by the cloud not slowing down too much the connection process in case of a real connection error.


src/ArduinoIoTCloudTCP.cpp
```diff
@@ -404,34 +406,58 @@ ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_SubscribeMqttTopics()
...
+    execCloudEventCallback(ArduinoIoTCloudEvent::CONNECT);
+
+    if (_shadowTopicIn != "")
+      return State::RequestLastValues;
+    else
+      return State::Connected;
+
+  }
+
+  return State::SubscribeMqttTopics;
```
polldo, line 460: I personally would write it in the following way to make it more readable (to avoid nested if), but it's just an opinion:
```C++
  if (!is_first_subscribe_request && !is_subscribe_retry_delay_expired)
  {
    return State::SubscribeMqttTopics;
  }

  if (_last_subscribe_request_cnt > AIOT_CONFIG_SUBSCRIBE_MAX_RETRY_CNT)
  {
    _last_subscribe_request_cnt = 0;
    _last_subscribe_request_tick = 0;
    _mqttClient.stop();
    execCloudEventCallback(ArduinoIoTCloudEvent::DISCONNECT);
    return State::ConnectPhy;
  }

  if (!_mqttClient.subscribe(_dataTopicIn))
  {
    DEBUG_ERROR("ArduinoIoTCloudTCP::%s could not subscribe to %s", __FUNCTION__, _dataTopicIn.c_str());
#if !defined(__AVR__)
    DEBUG_ERROR("Check your thing configuration, and press the reset button on your board.");
#endif
    _last_subscribe_request_tick = now;
    _last_subscribe_request_cnt++;
    return State::SubscribeMqttTopics;
  }

  if (_shadowTopicIn != "")
  {
    if (!_mqttClient.subscribe(_shadowTopicIn))
    {
      DEBUG_ERROR("ArduinoIoTCloudTCP::%s could not subscribe to %s", __FUNCTION__, _shadowTopicIn.c_str());
#if !defined(__AVR__)
      DEBUG_ERROR("Check your thing configuration, and press the reset button on your board.");
#endif
      _last_subscribe_request_tick = now;
      _last_subscribe_request_cnt++;
      return State::SubscribeMqttTopics;
    }
  }

  DEBUG_INFO("Connected to Arduino IoT Cloud");
  execCloudEventCallback(ArduinoIoTCloudEvent::CONNECT);
  _last_subscribe_request_cnt = 0;
  _last_subscribe_request_tick = 0;
  if (_shadowTopicIn != "")
  {
    return State::RequestLastValues;
  }
  return State::Connected;
```

polldo (approved):
lgtm, I left just an advice

src/ArduinoIoTCloudTCP.cpp
aentinger: I second @polldo's comment.

pennam: Thanks @polldo:
 - nested if: i would leave this one as it is 
```
   if (_shadowTopicIn != "")
   {
     if (!_mqttClient.subscribe(_shadowTopicIn))
```
since it wi be removed when we will merge the thing_id_discovery PR #297 
 - which is the advantage moving this part inside the ifs? Personally i would leave it outside.
```
 _last_subscribe_request_tick = now;
 _last_subscribe_request_cnt++;
```

polldo: >Personally i would leave it outside.

do it! 
I moved them to make it clear that those statements depended on the subscription fail

PR #307, opened 2022-03-01
ECCX08Cert and CryptoUtil: arrangements for the introduction of a new hardware crypto

pennam:
Goal of this PR is to ease the introduction of a new hardware crypto device in the ArduinoIoTCloud stack:

*  `ECCX08CertClass` is substituted with a more generic `ArduinoIoTCloudCertClass` based on the original `ECCX08CertClass` removing all references to `ECCX08` device
* CryptoUtilClass is extended in order to abstract basic crypto API and implementing function to build the Certificate Signing Request and the Certificate Itself.

~~I've forgot to fix the provisioning example :crying_cat_face:~~

|   | Provisioning | Connection  |
|--|--|--|
| MKR WIFI 1010  |  :green_heart: | :green_heart:  |
| NANO RP2040  |  :green_heart: | :green_heart:  |
| NANO 33 IOT  |  :green_heart: | :green_heart:  |
| PORTENTA   |  :green_heart: | :green_heart:  |
| MKR GSM 1400   |  :green_heart: | :green_heart:  |

PR #308, opened 2022-03-03
Prepare library to support SE050 crypto device

pennam:
This PR should be rebased and applied after #307 so only commits from https://github.com/arduino-libraries/ArduinoIoTCloud/commit/44a5920ddbd07c796397d7d1526df02d8d59830f to https://github.com/arduino-libraries/ArduinoIoTCloud/commit/3d79839629d963e80ce09cc5bc39c2a6b9562684 are meaningful. I did not found a better way to stack PRs from a fork...

Its goal is to prepare the library and add support to the `SE050` crypto device.

As of now this `SE050` can be found on `PORTENTA H7` boards, but it is not yet enabled in AIoTC_Config.h neither supported by the core.

### DIFF
--- a/src/tls/AIoTCSSCert.h
+++ b/src/tls/AIoTCSSCert.h
@@ -1,0 +1,48 @@
+/*
+   This file is part of ArduinoIoTBearSSL.
+
+   Copyright 2019 ARDUINO SA (http://www.arduino.cc/)
+
+   This software is released under the GNU General Public License version 3,
+   which covers the main part of ArduinoIoTBearSSL.
+   The terms of this license can be found at:
+   https://www.gnu.org/licenses/gpl-3.0.en.html
+
+   You can be released from the requirements of the above licenses by purchasing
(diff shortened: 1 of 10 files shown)


facchinm (approved):

src/tls/AIoTCSSCert.h
```diff
@@ -0,0 +1,48 @@
...
+ ******************************************************************************/
+
+#include <AIoTC_Config.h>
+#ifdef BOARD_HAS_SE050
+
+/******************************************************************************
+ * CONSTANTS
+ ******************************************************************************/
+static const char AIoTSSCert[] =
+"-----BEGIN CERTIFICATE-----\n"
```
aentinger, line 33: I suppose this is a **root** certificate for the Arduino IoT Cloud servers?

pennam: yes, it is. You can view it using:

`openssl s_client -connect server:port -showcerts`

PR #309, opened 2022-03-08
Add support for NICLA VISION

pennam:
This PR finalize the work done in #307 and #308 adding a new supported board


pennam:
CI fails because an old version of `Arduino_Portenta_OTA` is used. It should be updated in a while. Sketches are building fine locally.

PR #310, opened 2022-03-08
PORTENTA_H7_M4: Remove M4 core from supported build options

pennam:
With the actual configuration of the library OTA is not enabled for PORTENTA_H7_M4:

https://github.com/arduino-libraries/ArduinoIoTCloud/blob/45d5672c4473c9663720dc1f187c03bffce46c6e/src/AIoTC_Config.h#L105-L115

so is it safe to not compile `portenta_h7_onOTARequest()` function for the M4 build.

This is also a possible fix for the compile-examples failure of the last release caused by restricting the support of Arduino_Portenta_OTA only to the M7 core:

https://github.com/arduino-libraries/Arduino_Portenta_OTA/blob/0652134e737b67575e2d1c41cb9fac238922ebd0/src/Arduino_Portenta_OTA_Config.h#L27-L31

### DIFF
--- a/.github/workflows/compile-examples.yml
+++ b/.github/workflows/compile-examples.yml
@@ -50,8 +50,6 @@ jobs:
             type: gsm
           - fqbn: arduino:samd:mkrnb1500
             type: nb
-          - fqbn: arduino:mbed:envie_m4
-            type: mbed
           - fqbn: arduino:mbed:envie_m7
             type: mbed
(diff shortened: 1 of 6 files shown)


aentinger (approved):
LGTM :+1:

pennam:
:wave: @aentinger just to let you know that i've updated the PR and removed all M4 build options from the core and from the CI because at the moment the wifi module is not working on the M4 core.

PR #316, opened 2022-05-04
Remove ArduinoBearSSL as a CI build dependency.

aentinger:
The code formerly integrated in ArduinoBearSSL has long since been integrated with ArduinoIoTCloud library.


pennam:
@aentinger I've added a commit to fix Self provisioning example build. Esp32 build failure will be fixed with #326

aentinger:
:heavy_check_mark:

PR #317, opened 2022-05-11
TimeService: check connection status before network/ntp time request

pennam:
Background: a user reported a [crash](https://forum.arduino.cc/t/arduino-cloud-now-supports-esp32-boards/867845/2?u=eclipse1985) using ArduinoIoTCloud on ESP32 . 

After a bit of research we discovered the root cause was an early NTP request due to a property initialization into the setup function. In the [arduino-esp32](https://github.com/espressif/arduino-esp32) core the network stack is initialized when the `.begin()` function is called and in case of this library this happens at the first call of `ArduinoCloud.update()`

This PR slightly chage the behaviour of TimeSerivice adding a connection check before starting a remote time request to the connection handler or via NTP.

Doing in this way we avoid the crash reported on esp32 core and for the other platforms to submit time request that would return invalid values.


pennam:
TESTS:
ESP32 DEVKIT V1 :heavy_check_mark: 
MKRWiFi 1010: :heavy_check_mark: 
RP2040 CONNECT: :heavy_check_mark: 
MKGGSM1400: :heavy_check_mark: 
PORTENTA H7: :heavy_check_mark: 
NICLA VISION: :heavy_check_mark:

PR #302, opened 2022-02-03
Minor: TimeService debug and formatting

pennam:
Changes:
 - print timezone info only if DEBUG_VERBOSE is enabled
 - add braces to a single statement "if" to improve readability

PR #318, opened 2022-05-25
TimeService update/refactor

pennam:
After #317 and #302 i thought to cleanup TimeService class before proceeding with other developments.


pennam:
superseded by #353

PR #326, opened 2022-07-11
ESP32 core 2.0.4 CI fixes

pennam:
The last ESP32 core is enforcing -Wall -Werror=all compiler flags for `more` and `all` warning levels using https://github.com/espressif/arduino-esp32/blob/2.0.4/platform.txt#L105-L110 so a wrong initialization order or an unused variable will cause a build error in our CI


aentinger (approved):
LGTM :+1:

PR #342, opened 2022-11-08
Ethernet: examples and readme update

pennam:
Add EthernetConnectionHandler in library examples and the list of ethernet enabled boards in the readme.

Bonus: 
https://github.com/arduino-libraries/ArduinoIoTCloud/commit/0802445d458a84cebdb9ac306beae4cd0395544a Make Basic example build for all ESP32 boards

https://github.com/arduino-libraries/ArduinoIoTCloud/commit/7c4eb0938a463446974629938c3efc52bdfde5b9 Align header format of library examples

### DIFF
--- a/examples/ArduinoIoTCloud-Advanced/arduino_secrets.h
+++ b/examples/ArduinoIoTCloud-Advanced/arduino_secrets.h
@@ -35,3 +35,11 @@
   #define SECRET_LOGIN ""
   #define SECRET_PASS ""
 #endif
+
+/* Portenta H7 + Ethernet shield */
+#if defined(BOARD_HAS_ETHERNET)
+  #define SECRET_OPTIONAL_IP ""
+  #define SECRET_OPTIONAL_DNS ""
+  #define SECRET_OPTIONAL_GATEWAY ""
+  #define SECRET_OPTIONAL_NETMASK ""
+#endif
--- a/examples/ArduinoIoTCloud-Basic/ArduinoIoTCloud-Basic.ino
+++ b/examples/ArduinoIoTCloud-Basic/ArduinoIoTCloud-Basic.ino
@@ -6,11 +6,11 @@
   * When you flip the switch in the Cloud dashboard the onboard LED lights gets turned ON or OFF.
 
   IMPORTANT:
-  This sketch works with WiFi, GSM, NB and Lora enabled boards supported by Arduino IoT Cloud.
+  This sketch works with WiFi, GSM, NB, Ethernet and Lora enabled boards supported by Arduino IoT Cloud.
   On a LoRa board, if it is configured as a class A device (default and preferred option), values from Cloud dashboard are received
   only after a value is sent to Cloud.
 
-  This sketch is compatible with:
+  The full list of compatible boards can be found here:
    - https://github.com/arduino-libraries/ArduinoIoTCloud#what
 */
 
--- a/examples/ArduinoIoTCloud-Basic/ArduinoIoTCloud-Basic.ino
+++ b/examples/ArduinoIoTCloud-Basic/ArduinoIoTCloud-Basic.ino
@@ -17,7 +17,7 @@
 #include "arduino_secrets.h"
 #include "thingProperties.h"
 
-#if defined(ESP32)
+#if !defined(LED_BUILTIN)
 static int const LED_BUILTIN = 2;
 #endif
 
--- a/examples/ArduinoIoTCloud-DeferredOTA/thingProperties.h
+++ b/examples/ArduinoIoTCloud-DeferredOTA/thingProperties.h
@@ -2,6 +2,7 @@
 #elif defined(BOARD_HAS_GSM)
 #elif defined(BOARD_HAS_LORA)
 #elif defined(BOARD_HAS_NB)
+#elif defined(BOARD_HAS_ETHERNET)
 #else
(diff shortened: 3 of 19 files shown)


examples/ArduinoIoTCloud-Advanced/arduino_secrets.h
```diff
@@ -35,3 +35,11 @@
   #define SECRET_LOGIN ""
   #define SECRET_PASS ""
 #endif
+
+/* Portenta H7 + Ethernet shield */
+#if defined(BOARD_HAS_ETHERNET)
+  #define SECRET_OPTIONAL_IP ""
```
aentinger, line 41: Why are those secrets called `_OPTIONAL_`?

examples/ArduinoIoTCloud-Basic/ArduinoIoTCloud-Basic.ino
```diff
@@ -6,18 +6,18 @@
...
-  This sketch is compatible with:
+  The full list of compatible boards can be found here:
    - https://github.com/arduino-libraries/ArduinoIoTCloud#what
 */
 
 #include "arduino_secrets.h"
 #include "thingProperties.h"
 
-#if defined(ESP32)
+#if !defined(LED_BUILTIN)
```
aentinger, line 20: Yes, that's much better :rofl: I'm afraid I was the author of the original implementation =)

examples/ArduinoIoTCloud-DeferredOTA/thingProperties.h
```diff
@@ -34,4 +35,9 @@ void initProperties() {
   LoRaConnectionHandler ArduinoIoTPreferredConnection(SECRET_APP_EUI, SECRET_APP_KEY, _lora_band::EU868, NULL, _lora_class::CLASS_A);
 #elif defined(BOARD_HAS_NB)
   NBConnectionHandler ArduinoIoTPreferredConnection(SECRET_PIN, SECRET_APN, SECRET_LOGIN, SECRET_PASS);
+#elif defined(BOARD_HAS_ETHERNET)
+  /* DHCP mode */
+  //EthernetConnectionHandler ArduinoIoTPreferredConnection;
+  /* Manual mode. It will fallback in DHCP mode if SECRET_OPTIONAL_IP is invalid or equal to "0.0.0.0" */
```
aentinger, line 41: Okay, the parameter is optional when using DHCP. But maybe a better naming would be `SECRET_MANUAL_IP`. Also its not really a secret, just a configuration item. Maybe `ETHERNET_MANUAL_CONFIG_IP`?

aentinger (approved):
LGTM :+1: Just some thought impulses re naming.

pennam:
Thanks @aentinger . I've used the same naming used when exporting a sketch from the cloud editor.

This defines are indeed not secrets, but it wasn't possible to remove the `SECRET_` because cloud side are handled as if they were. The `_OPTIONAL_` part was added because the user may want or not to customize ethernet parameters. If it wont customize them the interface will work in DHCP mode.

Another :thinking:  fact is they are defined into the arduino_secrets.h file and the filename is not really appropriate in case of ethernet configuration parameter.

I feel like renaming only, would not be the best solution: maybe we should also move them in the thingProperties.h file. What do you think?

```
#if defined(BOARD_HAS_ETHERNET)
  /* DHCP mode */
  //EthernetConnectionHandler ArduinoIoTPreferredConnection;
  /* Manual mode. It will fallback in DHCP mode if SECRET_OPTIONAL_IP is invalid or equal to "0.0.0.0" */
  #define ETHERNET_MANUAL_CONFIG_IP ""
  #define ETHERNET_MANUAL_CONFIG_DNS ""
  #define ETHERNET_MANUAL_CONFIG_GATEWAY ""
  #define ETHERNET_MANUAL_CONFIG_NETMASK ""
  EthernetConnectionHandler ArduinoIoTPreferredConnection(ETHERNET_MANUAL_CONFIG_IP, ETHERNET_MANUAL_CONFIG_DNS, ETHERNET_MANUAL_CONFIG_GATEWAY, ETHERNET_MANUAL_CONFIG_NETMASK);
#elif defined(BOARD_HAS_WIFI)
  WiFiConnectionHandler ArduinoIoTPreferredConnection(SECRET_SSID, SECRET_PASS);
#elif defined(BOARD_HAS_GSM)
  GSMConnectionHandler ArduinoIoTPreferredConnection(SECRET_PIN, SECRET_APN, SECRET_LOGIN, SECRET_PASS);
#elif defined(BOARD_HAS_LORA)
  LoRaConnectionHandler ArduinoIoTPreferredConnection(SECRET_APP_EUI, SECRET_APP_KEY, _lora_band::EU868, NULL, _lora_class::CLASS_A);
#elif defined(BOARD_HAS_NB)
  NBConnectionHandler ArduinoIoTPreferredConnection(SECRET_PIN, SECRET_APN, SECRET_LOGIN, SECRET_PASS);
#endif
```

examples/ArduinoIoTCloud-DeferredOTA/thingProperties.h
aentinger: Thank you for your [clarification](https://github.com/arduino-libraries/ArduinoIoTCloud/pull/342#issuecomment-1306877015). I thought so. In this case there's not much you do I suppose :shrug:.

PR #343, opened 2022-11-14
Reorder and simplify addPropertyReal methods 

pennam:
During the `thing_id` discovery work https://github.com/arduino-libraries/ArduinoIoTCloud/pull/297 i've added unnecessary methods to the public space.

The goal of this pr is to:
reorder `addPropertyReal` methods in ArduinoIoTCloud.cpp to make more clear who is calling who
remove unnecessary `addPropertyReal` methods from public space: the user should not be able to select the property container because all user properties have to be added to the `_thing_property_container`.

### DIFF
--- a/src/ArduinoIoTCloud.cpp
+++ b/src/ArduinoIoTCloud.cpp
@@ -205,32 +205,6 @@ __attribute__((weak)) void setDebugMessageLevel(int const /* level */)
   /* do nothing */
 }
 
-/* The following methods are used internally to handle hidden thing and device properties */
-Property& ArduinoIoTCloudClass::addPropertyReal(bool& property, PropertyContainer &prop_cont, String name, Permission const permission)
-{
-  return addPropertyReal(property, prop_cont, name, -1, permission);
-}
-Property& ArduinoIoTCloudClass::addPropertyReal(float& property, PropertyContainer &prop_cont, String name, Permission const permission)
-{
-  return addPropertyReal(property, prop_cont, name, -1, permission);
-}
-Property& ArduinoIoTCloudClass::addPropertyReal(int& property, PropertyContainer &prop_cont, String name, Permission const permission)
-{
-  return addPropertyReal(property, prop_cont, name, -1, permission);
-}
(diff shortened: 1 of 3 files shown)


src/property/types/CloudString.h
```diff
@@ -49,6 +49,12 @@ class CloudString : public Property {
     void clear() {
       _value = PropertyActions::CLEAR;
     }
+    const char * c_str() {
+      return _value.c_str();
+    }
+    String& getValue() {
```
aentinger, line 55: Maybe a const reference `String const &` ? Otherwise you could and up with a unplanned modification of the underlying string.

src/property/types/CloudString.h
```diff
@@ -49,6 +49,12 @@ class CloudString : public Property {
     void clear() {
       _value = PropertyActions::CLEAR;
     }
+    const char * c_str() {
```
aentinger, line 52: Const-correctness might make sense.
```suggestion
    const char * c_str() const {
```

aentinger (approved):
Well, I'm a big fan of any kind of API surface reduction, so kudos :wink:

pennam:
Thanks for your review @aentinger. You made me remember that after i've reverted https://github.com/arduino-libraries/ArduinoIoTCloud/commit/cd98ce21c117f5da1218fe0cbbb15843db059380 also the followings two commits are no more needed:
 - https://github.com/arduino-libraries/ArduinoIoTCloud/pull/343/commits/e8a5d7c4e8efbe7e17d6c6c6ae0ad225aa4b63dc
 - https://github.com/arduino-libraries/ArduinoIoTCloud/pull/343/commits/7ee9b6c0674664bd9140422cdbbf8488fad8d7c2
 
 So i think i will cleanup this PR dropping all that is not necessary :bow:

aentinger:
Glad that looking over your PR was helpful in some way =)

src/property/types/CloudString.h
pennam: this change was dropped

src/property/types/CloudString.h
pennam: this change was dropped

PR #344, opened 2022-11-14
Fix compilation warnings

pennam:
Just a round on compilation warnings

### DIFF
--- a/src/utility/time/NTPUtils.h
+++ b/src/utility/time/NTPUtils.h
@@ -54,7 +54,7 @@ private:
   static int           const MAX_NTP_PORT         = 65535;
 #endif
   static unsigned long const NTP_TIMEOUT_MS       = 1000;
-  static char constexpr *    NTP_TIME_SERVER      = "time.arduino.cc";
+  static constexpr const char * NTP_TIME_SERVER   = "time.arduino.cc";
 
   static void sendNTPpacket(UDP & udp);
 };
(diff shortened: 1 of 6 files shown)


src/utility/time/NTPUtils.h
```diff
@@ -54,7 +54,7 @@ class NTPUtils
   static int           const MAX_NTP_PORT         = 65535;
 #endif
   static unsigned long const NTP_TIMEOUT_MS       = 1000;
-  static char constexpr *    NTP_TIME_SERVER      = "time.arduino.cc";
+  static constexpr const char * NTP_TIME_SERVER   = "time.arduino.cc";
```
aentinger, line 57: I wonder if `constexpr const char` is not a bit redundant.

Should not `static constexpr char * NTP_TIME_SERVER = "time.arduino.cc";`be enough?

aentinger (approved):
LGTM :+1:

src/utility/time/NTPUtils.h
pennam: I was also thinking so, but after a search about the warning i've foud out that:

`constexpr const char * p ` declares a const pointer to const char data
`constexpr char * p ` declares a const pointer to char data

aentinger: Thank you for checking that up ;) Now I also grew a bit smarter =)

PR #350, opened 2023-01-18
Add esp32 OTA support

pennam:
Needs: https://github.com/bcmi-labs/Arduino_ESP32_OTA


pennam:
Magic number fixed Thanks @mirkokurt :rocket:

pennam:
CI fixed for mkr100 mkrwan and mkrnb. esp32 still fails becauseArduino_ESP32_OTA is not yet published.

extras/tools/bin2ota.py
```diff
@@ -26,6 +26,8 @@
     magic_number = 0x2341025B.to_bytes(4,byteorder='little')
 elif board == "NANO_RP2040_CONNECT":
     magic_number = 0x2341005E.to_bytes(4,byteorder='little')
+elif board == "ESP32":
+    magic_number = 0x45535033.to_bytes(4,byteorder='little')
```
aentinger, line 30: Is there the same VID/PID for all ESP32 boards? Somehow this is hard to believe, unless I'm missing something here.

aentinger (commented):
Looks good to me, love the refactoring in different source files for different architectures, much cleaner.

aentinger:
> CI fixed for mkr100 mkrwan and mkrnb. esp32 still fails becauseArduino_ESP32_OTA is not yet published.

You could make this one public: https://github.com/bcmi-labs/Arduino_ESP32_OTA and then specify the link for the CI action to work, i.e.

```diff
- name: Arduino_ESP32_OTA
+ - source-url: https://github.com/bcmi-labs/Arduino_ESP32_OTA
```

Of course you still need to update this once the library is publisher 🤷

extras/tools/bin2ota.py
pennam: You are correct, there is really a huge amount of VID/PID pairs in case of ESP boards. For this reason this is intended as a generic magic number not a VID/PID combination. I'll add a note to make it clear.

The Updater class however will check for the binary compatibility using the ESP binary header

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/app_image_format.html#structures

and checking if the `chip_id` , that differs from ESP32 to ESP32-S2 and ..., is matching.

aentinger: Roger that, thank you for the clarification ;)

PR #354, opened 2023-02-26
Add Arduino_ESP32_OTA to library dependencies

alranel:
This is probably a regression: compilation fails on ESP32 when the new Arduino_ESP32_OTA is not installed. This PR adds it to dependencies so that it's automatically installed.


per1234 (approved):

pennam (commented):
:+1:  Thanks @alranel

PR #380, opened 2023-09-27
Arduino IoT Cloud Docs

karlsoderby:
Markdown style documentation for this library (following our traditional approach). 
Docs will be rendered at https://www.arduino.cc/reference/en/libraries/arduinoiotcloud/


docs/api.md
```diff
@@ -0,0 +1,531 @@
...
+```
+ArduinoCloud.setTimestamp(cloudVariable, timestamp)
+```
+
+#### Parameters
+
+- `String` - property name
+- `unsigned long const` - timestamp
+
+### `setThingId()`
```
pennam, line 54: This funciton is only needed by LoRa devices, since for all TCP devices the ThingId is automaticallly configured by the Cloud via ThingId discovery protocol.

docs/api.md
```diff
@@ -0,0 +1,531 @@
...
+```
+ArduinoCloud.deviceNotAttached()
+```
+
+#### Parameters
+None.
+
+#### Returns
+
+- Boolean (true/false)
```
pennam, line 170: I would not add this functions to the docs since are not intented to be used by the end user and we are planning to remove them from public API

docs/api.md
```diff
@@ -0,0 +1,531 @@
...
+```
+
+#### Parameters
+None
+
+#### Returns
+- `unsigned long` - local time
+
+
+### `updateInternalTimezoneInfo()`
```
pennam, line 217: Same for this. I would really want to remove this from public API.

docs/api.md
```diff
@@ -0,0 +1,531 @@
...
+
+```
+ArduinoCloud.setBoardId(DEVICE_LOGIN_NAME);
+```
+
+### `setSecretDeviceKey()`
+
+#### Description
+
+This method is only enabled if you are using an ESP32/ESP8266 based board, and sets the secret device key for the device. The secret key is only obtainable from the Arduino IoT Cloud during device configuration.
```
pennam, line 348: Temporarily this is also used by the UNO R4, but it will change in the near future.

docs/api.md
```diff
@@ -0,0 +1,531 @@
...
+### `update()` (LPWAN)
+
+#### Description
+
+**Note:** for TCP devices, the `update()` function is implemented differently. See [ArduinoCloud Class (TCP)](). 
+
+Responsible for updating the IoT Cloud connection. 
+- First it feeds a watchdog timer, making sure it is not stuck anywhere,
+- then sends variable updates to the IoT Cloud (via MQTT client)
+- finally, it checks for new data from the IoT Cloud (via the MQTT client)
```
pennam, line 418: LoRa do not use MQTT protocol

pennam (changes requested):

pennam:
Thanks @karlsoderby i left some comments for you.

pennam:
Also, some of the boards listed here: https://www.arduino.cc/reference/en/libraries/arduinoiotcloud/ are not compatible with ArduinoIoTCloud and shloud be removed.

docs/api.md
karlsoderby: Ahh yes this was incorrectly added from the TCP method. Corrected!

docs/api.md
karlsoderby: Good call. I added the Nano ESP32 as well.

docs/api.md
karlsoderby: aha thanks for the clarification, it has been removed.

karlsoderby:
@pennam thanks for the review, I've addressed the changes just now. 

Regarding 

> Also, some of the boards listed here: https://www.arduino.cc/reference/en/libraries/arduinoiotcloud/ are not compatible with ArduinoIoTCloud and shloud be removed.

Unfortunately they are added based on the `architectures` field inside of the `library.properties` file. Any board that is part of our listed architectures are (incorrectly) listed. That's why boards like MKR Zero and Nano BLE is listed.

pennam (approved):

PR #408, opened 2024-01-16
OTA: making it possible to perform ota without the need of the thing ID

andreagilardoni:


src/ArduinoIoTCloudTCP.cpp
```diff
@@ -270,6 +270,12 @@ void ArduinoIoTCloudTCP::update()
   }
   _state = next_state;
 
+#if OTA_ENABLED
+  if(_state >= State::SubscribeDeviceTopic && _state <= State::Connected) {
```
pennam, line 274:
```suggestion
  if(_state > State::SubscribeDeviceTopic && _state <= State::Connected) {
```
If `State` is exactly `SubscribeDeviceTopic` you cannot be sure that the board is correctly subscribed to the device topic

pennam (changes requested):

PR #416, opened 2024-02-14
Increase MQTT buffer size on STM32H7

manchoz:
Increasing MQTT buffer size on STM32H7-based board allows use cases such as using large `String` values to carry payloads to 3rd platforms.


pennam (changes requested):
@manchoz do you mind adding a configuration define in https://github.com/arduino-libraries/ArduinoIoTCloud/blob/master/src/AIoTC_Config.h file ?

something like #define AIOT_CONFIG_MQTT_TX_BUFFER_SIZE

PR #425, opened 2024-02-16
ArduinoCloudTCP: Separate the logic of the device and the logic of the thing in two different state machines

mirkokurt:
This PR aims to separate the logic of the device (connect to the physical network, connect to MQTT, connect to device topics...) from the logic of the thing (receive the thing id, connect to the thing topics, sync and send telemetry) splitting the whole connection logic in two different state machines. 

To achieve this, a new ArduinoIoTCloudTCPThing class has been created as member of the ArduinoIoTCloudTCP class.
The new class implement a state machine dedicated to the thing logic. The device logic has been kept in the ArduinoIoTCloudTCP.

From the user perspective, the only difference is that now the device state machine is not blocked anymore waiting the thing id when the thing id is not provided by the backend. The device state machine is independent from the thing configuration and, theoretically, is able to perform an OTA even if not attached to a thing (theoretically because the backend does not allow this).

From the developer perspective, this refactor allows a better logical separation of behaviours, improved readability and maintainability, and it's preparatory to future enhancements of the OTA feature.


src/AIoTC_Types.h
```diff
@@ -0,0 +1,51 @@
...
+   You can be released from the requirements of the above licenses by purchasing
+   a commercial license. Buying such a license is mandatory if you want to modify or
+   otherwise use the software for commercial activities involving the Arduino
+   software without disclosing the source code of your own applications. To purchase
+   a commercial license, send an email to license@arduino.cc.
+*/
+
+#ifndef ARDUINO_AIOTC_TYPES_H_
+#define ARDUINO_AIOTC_TYPES_H_
+
```
andreagilardoni, line 20:
```suggestion

#include <stdint.h>

```

src/AIoTC_Types.h
```diff
@@ -0,0 +1,51 @@
...
+/******************************************************************************
+   TYPEDEF
+ ******************************************************************************/
+
+typedef enum
```
andreagilardoni, line 25:
```suggestion
typedef enum: uint8_t
```
We don't need and entire int32 to represent the R/W modes

PR #428, opened 2024-02-19
chore: Switch LPWAN off deprecated method

zfields:


pennam:
Thanks for your contribution @zfields

pennam:
Hi @zfields can you add a bit a context about the need to change `_intervalRetry`? I was thinkiong to add a configuration define inside  [AIoTC_Config.h](https://github.com/arduino-libraries/ArduinoIoTCloud/blob/master/src/AIoTC_Config.h) . What do you think about it?

zfields:
REASON: In my testing with the MKR 1310, I noticed the modem (`MKRWAN`) would return a busy error if request came too much faster than once every 10 seconds. This is an empirical observation more than science.

I think a config define is a great idea.

zfields:
@pennam, I have implemented the update you suggested.

zfields:
If you don't want this change, will you let me know so I can delete and stop maintaining the branch?

Or if it needs to be fixed, just let me know what you would like to see change.

pennam:
> @pennam If you don't want this change, will you let me know so I can delete it and stop maintaining the branch?
> 
> Or if it needs to be fixed, just let me know what you would like to see changed.

hi @zak sorry that this PR is taking more than expected to get merged. I'll do it today :) Just wanted to give it a spin to test and understand if and how the increased retry delay is affecting the communication.

btw you don't need to force push you PRs to keep them aligned to master unless there are conflicts.

Thanks

PR #434, opened 2024-02-28
Update ArduinoIoTCloud.h

kindofn0pe:
Fixed conflict issue with NimBLE library where READ and WRITE are already defined


pennam:
@kindofn0pe Thanks for your contribution, unfortunately i cannot merge this PR because will break builds of Cloud generated sketches.

PR #435, opened 2024-03-12
fix: Add dependency: `Arduino_SecureElement`

zfields:
Recent change takes a dependency on `Arduino_SecureElement`, but failes to include the dependency in `library.properties`.


pennam:
:rocket: Thanks @zfields

PR #450, opened 2024-04-22
Examples: Avoid SECRET_PASS redefinition on Portenta H7

pennam:


per1234 (changes requested):
There is one more occurrence of the problem here:

https://github.com/arduino-libraries/ArduinoIoTCloud/blob/944f4f267639ab405b7e981ae66e3535f93b2369/examples/utility/ArduinoIoTCloud_Travis_CI/arduino_secrets.h#L8

pennam:
@per1234 i've force pushed the fix. Thanks

per1234 (approved):
Thanks Mattia!

PR #447, opened 2024-04-22
Add new Command topics

andreagilardoni:

PR #455, opened 2024-05-09
Fix thing reconfiguration after wifi connection loss

pennam:


pennam:
Manually merged here https://github.com/arduino-libraries/ArduinoIoTCloud/pull/447/commits/d755efe5ee966adad95974e814b1de9f47f014f0

PR #472, opened 2024-06-03
Watchdog: use mbed API

pennam:
This is needed to allow ArduinoCellular library to kick the watchdog without using callback


andreagilardoni (approved):

pennam:
Force pushed to fix build on ARDUINO NANO RP2040

PR #476, opened 2024-06-04
OTA: increase download time to 2s

pennam:
Apply after #475

### DIFF
--- a/src/ota/interface/OTAInterfaceDefault.cpp
+++ b/src/ota/interface/OTAInterfaceDefault.cpp
@@ -94,7 +94,9 @@ OTACloudProcessInterface::State OTADefaultCloudProcessInterface::fetch() {
 
   do {
     if(http_client->available() == 0) {
-      goto exit;
+      /* Avoid tight loop and allow yield */
+      delay(1);
+      continue;
     }
 
     http_res = http_client->read(context->buffer, context->buf_len);
--- a/src/ota/interface/OTAInterfaceDefault.h
+++ b/src/ota/interface/OTAInterfaceDefault.h
@@ -51,7 +51,7 @@ private:
 
   // The amount of time that each iteration of Fetch has to take at least
   // This mitigate the issues arising from tasks run in main loop that are using all the computing time
-  static constexpr uint32_t downloadTime = 100;
+  static constexpr uint32_t downloadTime = 2000;
 
   enum OTADownloadState: uint8_t {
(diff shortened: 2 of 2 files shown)


src/ota/interface/OTAInterfaceDefault.cpp
```diff
@@ -94,19 +93,26 @@ OTACloudProcessInterface::State OTADefaultCloudProcessInterface::fetch() {
...
-      goto exit;
-    }
+    if(http_client->available()) {
+      http_res = http_client->read(context->buffer, context->buf_len);
+
+      if(http_res < 0) {
+        DEBUG_VERBOSE("OTA ERROR: Download read error %d", http_res);
+        res = OtaDownloadFail;
+        goto exit;
+      }
```
andreagilardoni, line 103:
```suggestion
    if(http_client->available() == 0) {
      /* Avoid tight loop and allow yield */
      delay(1);
      continue;
    }

    http_res = http_client->read(context->buffer, context->buf_len);

    if(http_res < 0) {
      DEBUG_VERBOSE("OTA ERROR: Download read error %d", http_res);
      res = OtaDownloadFail;
      goto exit;
    }
```

src/ota/interface/OTAInterfaceDefault.cpp
```diff
@@ -94,19 +93,26 @@ OTACloudProcessInterface::State OTADefaultCloudProcessInterface::fetch() {
...
-      res = OtaDownloadFail;
-      goto exit;
+      if(context->writeError) {
+        DEBUG_VERBOSE("OTA ERROR: File write error");
+        res = ErrorWriteUpdateFileFail;
+        goto exit;
+      }
+    } else {
+      /* Avoid tight loop and allow yield */
+      delay(1);
```
andreagilardoni, line 114: This change is a consequence of the previous comment

```suggestion
    if(context->writeError) {
      DEBUG_VERBOSE("OTA ERROR: File write error");
      res = ErrorWriteUpdateFileFail;
      goto exit;
    }
```

src/ota/interface/OTAInterfaceDefault.cpp
pennam: personally i think the `if else` solution is more intuitive, but i also think a `continue` will feel good in this `goto` function :smile:

andreagilardoni (approved):

PR #485, opened 2024-07-02
fix: infinite disconnect loop

zfields:
When a disconnect occurs, a reset message is provided to both the `_thing` and `_device` containers, but there is no subsequent call to `_thing.update()` or `_device.update()` to process the message and actually reset the state.

This results in an infinite loop, because `!_device.connected()` (i.e. not not disconnected) will ALWAYS remain `true`.

This seems like the appropriate place for the call to `update()`, because `ResetCmdId` puts the `_device` and `_thing` containers into their disconnected state then `update()` processes the disconnect and restores them to their initialization state (same state that results from `begin()` just before we enter the state machine).

### DIFF
--- a/src/ArduinoIoTCloudTCP.cpp
+++ b/src/ArduinoIoTCloudTCP.cpp
@@ -346,9 +346,14 @@ ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_Disconnect()
     _mqttClient.stop();
   }
 
+  // Reset the Thing property container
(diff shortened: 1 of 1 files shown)


pennam:
Thanks for the headsup @zfields . I would fix this with #490 . Do you have time to give it a spin?

zfields:
Yes. I'd be happy to. I'm testing now, so it won't be a problem.

pennam:
superseded by #490

PR #488, opened 2024-07-16
moving mqtt poll into ArduinoIoTCloudTCP::handle_Connected

andreagilardoni:


src/ArduinoIoTCloudTCP.cpp
```diff
@@ -307,6 +303,7 @@ ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_Connected()
     return State::Disconnect;
   }
 
+  _mqttClient.poll();
```
pennam, line 306:
```suggestion
  /* Check for new data from the MQTT client. */
  _mqttClient.poll();
```

pennam (changes requested):
@andreagilardoni can you keep also the comment

src/ArduinoIoTCloudTCP.cpp
```diff
@@ -307,6 +303,8 @@ ArduinoIoTCloudTCP::State ArduinoIoTCloudTCP::handle_Connected()
     return State::Disconnect;
   }
 
+  /* Check for new data from the MQTT client. */
+  _mqttClient.poll();
```
pennam, line 307:
```suggestion
  _mqttClient.poll();

```

pennam (approved):

PR #490, opened 2024-07-18
Improve device and thing state machine message handling and fix infinite disconnetct loop

pennam:


zfields:
I have this patch on hardware, and I will let you know how it performs.

zfields:
You patch works on my application. :tada:

zfields (approved):
This change causes the correct behavior in my application.

PR #504, opened 2024-08-22
Cmake improvements

andreagilardoni:


pennam:
Any chance to use https://github.com/catchorg/Catch2/releases/download/v3.4.0/catch_amalgamated.hpp instead of the whole library? Using this PR it takes more time to compile the tests becase of increased build time related to Catch2 library.

`cmake-improvements` test build time 53s
`master` test build time 26s

andreagilardoni:
In [this link](https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md#how-to-migrate-projects-from-v2-to-v3) they suggest using the cmake approach, claiming that in the new version cmake integration is faster in terms of compiling time, wrt amalgamated version.

I run the tests on the following branches with:
```
cmake .
time make
```
Branch | Time
---|---
master | 29.047s
cmake-improvements | 59.656s
cmake-improvements-amalgamated | 39.765s

cmake-improvements-amalgamated requires to drop Cpp11 standard in favor for a newer one otherwise it won't compile. It seems that in the non amalgamated version we are not required to use cpp14, for some reasons.

I would personally prefer the `cmake-improvements` approach because he management of the dependency relies on cmake, instead of manual copy and paste of files from another repository. 
Probably we still need to improve includes from test sources to perform better.

pennam (approved):

pennam (approved):

pennam (approved):

PR #507, opened 2024-08-23
Fix time validity check to account for maximum possible timezone offset

### LINKED ISSUE #506: "ArduinoIoTCloudTCP::handle_SyncTime could not get valid time" caused by timezone diffrence
If the user's timezone is higher than UTC (ex. UTC+4) than if the code is compiled at `22.00 UTC`, `EPOCH_AT_COMPILE_TIME` will return a value higher than the current epoch because `cvt_time(__DATE__)` returns the date according to the user's local timezone.

This means that code compiled at `23.00 UTC` will not function for timezones `UTC+1` and higher.

[according to wiki](https://en.wikipedia.org/wiki/Time_zone) the biggest offset is `UTC+14` so decreasing `50400` (14 * 60 * 60) would work:

[ArduinoIoTCloud/src/utility/time/TimeService.cpp line 97](https://github.com/arduino-libraries/ArduinoIoTCloud/blob/3dfe573cd34333e4d1848eec0df3e2ee2f53dc92/src/utility/time/TimeService.cpp#L97)
```cpp
#define EPOCH_AT_COMPILE_TIME (cvt_time(__DATE__) - 50400)
```

or

[ArduinoIoTCloud/src/utility/time/TimeService.cpp line 323](https://github.com/arduino-libraries/ArduinoIoTCloud/blob/3dfe573cd34333e4d1848eec0df3e2ee2f53dc92/src/utility/time/TimeService.cpp#L323)
```cpp
bool TimeServiceClass::isTimeValid(unsigned long const time)
{
  return (time > (EPOCH_AT_COMPILE_TIME - 50400));
}
```

This may not be the best solution but is one nevertheless.
I can open a simple PR for this if it is ok for you guys.

kytpbs:
Fixes #506

### DIFF
--- a/src/utility/time/TimeService.cpp
+++ b/src/utility/time/TimeService.cpp
@@ -320,7 +320,9 @@ unsigned long TimeServiceClass::getRemoteTime()
 
 bool TimeServiceClass::isTimeValid(unsigned long const time)
 {
-  return (time > EPOCH_AT_COMPILE_TIME);
+  // EPOCH_AT_COMPILE_TIME is in local time, 
+  // so we need to subtract the maximum possible timezone offset to make sure we are less then utc time
+  return (time > (EPOCH_AT_COMPILE_TIME - (/*UTC+14*/ 14 * 60 * 60)));
 }
 
 bool TimeServiceClass::isTimeZoneOffsetValid(long const offset)
--- a/src/utility/time/TimeService.cpp
+++ b/src/utility/time/TimeService.cpp
@@ -320,9 +320,10 @@ unsigned long TimeServiceClass::getRemoteTime()
 
 bool TimeServiceClass::isTimeValid(unsigned long const time)
 {
-  // EPOCH_AT_COMPILE_TIME is in local time, 
-  // so we need to subtract the maximum possible timezone offset to make sure we are less then utc time
-  return (time > (EPOCH_AT_COMPILE_TIME - (/*UTC+14*/ 14 * 60 * 60)));
+  /* EPOCH_AT_COMPILE_TIME is in local time, so we need to subtract the maximum
+   * possible timezone offset UTC+14 to make sure we are less then UTC time
+   */
+  return (time > (EPOCH_AT_COMPILE_TIME - (14 * 60 * 60)));
 }
 
 bool TimeServiceClass::isTimeZoneOffsetValid(long const offset)


kytpbs:
@pennam could you review the PR? Would be very appreciated.

src/utility/time/TimeService.cpp
```diff
@@ -320,7 +320,9 @@ unsigned long TimeServiceClass::getRemoteTime()
 
 bool TimeServiceClass::isTimeValid(unsigned long const time)
 {
-  return (time > EPOCH_AT_COMPILE_TIME);
+  // EPOCH_AT_COMPILE_TIME is in local time, 
+  // so we need to subtract the maximum possible timezone offset to make sure we are less then utc time
+  return (time > (EPOCH_AT_COMPILE_TIME - (/*UTC+14*/ 14 * 60 * 60)));
```
pennam, line 325:
```suggestion
  /* EPOCH_AT_COMPILE_TIME is in local time, so we need to subtract the maximum
   * possible timezone offset UTC+14 to make sure we are less then UTC time
   */
  return (time > (EPOCH_AT_COMPILE_TIME - (14 * 60 * 60)));
```

pennam (changes requested):
Hi @kytpbs,
thanks for your contribution and sorry for the delay I was OOF the past two weeks. I've got only one style change request. I suggested you the changes in order to match the comment style of the rest of the file.

Thanks

pennam:
Thanks, tomorrow morning i will give this PR the last spin and then i will merge it.

pennam:
Hi @kytpbs I've made some tests and found a glitch with this changes:

`EPOCH_AT_COMPILE_TIME` is now considered a valid time and thus it will be stored inside the RTC instead of skipping it and retry to get a valid time.

I'm working to fix this regression and will propose you a couple of commits to be integrated in this PR.

pennam:
Hi @kytpbs I've prepared this two commits if you want to cherry-pick them and doublecheck your issue is fixed.

https://github.com/pennam/ArduinoIoTCloud/commit/419b39baae50bcfb2ec62002465850ab941c7347
https://github.com/pennam/ArduinoIoTCloud/commit/012b73e65da49298493aa70f64a9f00dcf996ae6

kytpbs:
at these [lines](https://github.com/pennam/ArduinoIoTCloud/commit/012b73e65da49298493aa70f64a9f00dcf996ae6#diff-6c5ff72bf95a5ae6b9c05d69677ba6f2c95c60e88f67e99dbe1e1ff2ea0849e5L172-L178) 
```cpp
#ifdef HAS_TCP
    utc = getRemoteTime();
#endif
#ifdef HAS_LORA
    /* Just keep incrementing stored RTC value*/
    /* Just keep incrementing stored RTC value starting from EPOCH_AT_COMPILE_TIME */
    utc = getRTC();
#endif
```

shouldn't we first check if we have RTC and use that and return or vice versa, since if `HAS_TCP` and `HAS_LORA` are both defined (for example ESP32), we will first get the remote time, just to overwrite it to `getRTC`? I know this is not part of the commits but it peaked my interest looking at the diffs

pennam:
Hi @kytpbs thanks for checking! None of the supported boards has both `HAS_TCP` and `HAS_LORA` defined. The sync method is intended as a way to keep in sync the RTC with some other more reliable time source:

If user has defined a custom sync function, this method will be used otherwise if baord `HAS_TCP` we try to get time from NTP of connection handler. Afaik there is no (simple) way to get a timestamp from LoRa network so we rely on the RTC value that at boot time would be configured to `EPOCH_AT_COMPILE_TIME`.

If in the future we will have a board with both TCP and LoRa probably it would be better to rewrite the function like this to avoid overwriting the value:

```
#ifdef HAS_TCP
    utc = getRemoteTime();
#else
    /* Just keep incrementing stored RTC value starting from EPOCH_AT_COMPILE_TIME */
    utc = getRTC();
#endif
```

kytpbs:
Thanks for the explanation

I will double-check if the issue is fixed tonight if I can since it only occurs after 12 AM

what should I do with your commits/branch
- cherry-pick add them to this branch
- merge that branch into this one
- do nothing since you will just merge your branch which also has my commits instead?

kytpbs:
Just tested my branch and things worked correctly

pennam:
@kytpbs Thanks for the test! Please cherry-pick the commits on this branch so we can better keep track of this PR history and related issue.

pennam:
Superseded by #509

PR #508, opened 2024-09-12
Notecard OTA

zfields:


zfields:
Currently, this compiles but will not properly manage the OTA of the UnoR4 or MKR boards.

PR #525, opened 2025-01-27
Print library version inside debug info

pennam:
Simple change to add useful information about library version


andreagilardoni (approved):

pennam:
@zfields are you ok with this change?

PR #532, opened 2025-02-19
TCP: do not disconnect if ota client is connected

pennam:
It can happen that mqttclient disconnects during ota fetching. In this case we should also check for ota client state before go in disconnected state.

If ota client is still connected is better to give ota priority and try to finish to download before disconnect.


andreagilardoni:
Should we instead move this logic inside of `ArduinoIoTCloudTCP::update()`?
- In this way the ota process is independent from the mqtt client and is able to continue, even if the client is disconnected. 
- The Mqtt client, in the mean time, is able to be restored
- If the failure is at network level both will fail and the ota FSM handles it by itself
- ota are triggered by mqtt incoming commands, so nothing strange will happen on that case

What do you think about it?

andreagilardoni:
I think that moving the call for `_ota.update();` could be just enough

andreagilardoni (approved):

PR #539, opened 2025-04-07
Explicitly handle NAN values

sebromero:
In the current version NAN values are not handled. This leads to unexpected behaviour since the numerical values of both the local variable and the cloud version are used to determine if there is a difference. Without this change a local variable that is initialized with a default of NAN will never receive the cloud value if the permission is set to `Write`. If the permission is set to `ReadWrite` the cloud value will always be overridden with NAN even with `onSync(CLOUD_WINS)`.

This PR explicitly handles the cases where either one of the values is NAN (non-equal), or both are NAN (equal).

### DIFF
--- a/extras/test/CMakeLists.txt
+++ b/extras/test/CMakeLists.txt
@@ -66,6 +66,7 @@ set(TEST_SRCS
   src/test_addPropertyReal.cpp
   src/test_callback.cpp
   src/test_CloudColor.cpp
+  src/test_CloudFloat.cpp
   src/test_CloudLocation.cpp
   src/test_CloudSchedule.cpp
   src/test_decode.cpp
--- a/extras/test/CMakeLists.txt
+++ b/extras/test/CMakeLists.txt
@@ -67,6 +67,7 @@ set(TEST_SRCS
   src/test_callback.cpp
   src/test_CloudColor.cpp
   src/test_CloudFloat.cpp
+  src/test_CloudWrapperFloat.cpp
   src/test_CloudLocation.cpp
   src/test_CloudSchedule.cpp
   src/test_decode.cpp
--- a/extras/test/src/test_CloudFloat.cpp
+++ b/extras/test/src/test_CloudFloat.cpp
@@ -1,0 +1,112 @@
+/*
+   Copyright (c) 2019 Arduino.  All rights reserved.
+*/
+
+/**************************************************************************************
+   INCLUDE
+ **************************************************************************************/
+
+#include <catch2/catch_test_macros.hpp>
+
+#include <CBORDecoder.h>
+
+#include <property/types/CloudFloat.h>
+
+/**************************************************************************************
+  TEST CODE
+ **************************************************************************************/
+
+SCENARIO("Arduino Cloud Properties ", "[ArduinoCloudThing::CloudFloat]")
+{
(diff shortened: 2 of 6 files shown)


sebromero:
@andreagilardoni NaN is a valid float value and the current implementation of isDifferentFromCloud yields incorrect behavior in presence of this value. I don't see how that could be desirable.

andreagilardoni:
@sebromero can you describe to me the incorrect behavior we have? and how can I replicate it?

sebromero:
Ciao @andreagilardoni ! Sure, I'll explain. In any case I don't understand what you mean by floating point exceptions propagating to cloud? This affected function just returns a boolean and Arduino has no exception handling, so I'm not sure what would propagate anywhere 🤔 If you mean that a local NAN value should propagate to the cloud, then I totally agree and this is exactly what this PR fixes.

So, the job of `isDifferentFromCloud` is to return true whenever the local value is not the same as in the cloud, no matter what value it is. NAN is a legit value, so it should be treated the same way. I'll formulate the following as a unit test, but for now just run this on your machine and you will see that the expected output doesn't match the actual output for the last four cases. It doesn't work for two reasons, we use `abs` instead of `fabs` for float numbers, and NAN != NAN. 

```cpp
#include <iostream>
#include <cmath>

bool isDifferentFromCloud(float _primitive_value, float _cloud_value, float _min_delta_property = 0.0f) {
    bool isEqual = _primitive_value == _cloud_value;
    std::cout << "Is equal " << _primitive_value << " == " << _cloud_value << ": " << isEqual << std::endl;
    int difference = abs(_primitive_value - _cloud_value);
    std::cout << "Difference " << _primitive_value << " - " << _cloud_value << ": " << difference << std::endl;
    bool isDifferent = _primitive_value != _cloud_value && difference >= _min_delta_property;
    std::cout << "Is different " << _primitive_value << " != " << _cloud_value << ": " << isDifferent << std::endl;
    std::cout << "-------------------" << std::endl;
    return isDifferent;
    // Source: return _primitive_value != _cloud_value && (abs(_primitive_value - _cloud_value) >= Property::_min_delta_property);
}

int main(){
    // Correct
    isDifferentFromCloud(1.5f, 1.0f); // Expected result: true, actual result: true
    isDifferentFromCloud(1.0f, 1.5f); // Expected result: true, actual result: true
    isDifferentFromCloud(1.5f, 1.5f); // Expected result: false, actual result: false
    isDifferentFromCloud(NAN, NAN); // Expected result: false, actual result: false (only because NAN != NAN)

    // Broken
    isDifferentFromCloud(1.5f, 1.0f, 0.25f); // Expected result: true, actual result: false
    isDifferentFromCloud(1.0f, 1.5f, 0.25f); // Expected result: true, actual result: false
    isDifferentFromCloud(NAN, 1.5f); // Expected result: true, actual result: false
    isDifferentFromCloud(1.5f, NAN); // Expected result: true, actual result: false
    return 0;
}
```

sebromero:
[Test.zip](https://github.com/user-attachments/files/19798580/Test.zip)
Looks good now. You can use the test in the attachment. Before this PR: When you set the variable to a number through the dashboard it would be overwritten to NaN when the board restarts. After the PR: The value from the cloud persists when the board restarts.

andreagilardoni (approved):

PR #553, opened 2025-06-03
Send Network configuration to Cloud

fabik111:
Changes:

- added new CBOR payload command `DeviceNetConfigCmdUp` for transferring the network configuration
- send the new command after the `DeviceBeginCmdId` comand

This PR relies on https://github.com/arduino-libraries/Arduino_ConnectionHandler/pull/143


src/cbor/CBOR.h
```diff
@@ -22,17 +22,18 @@
...
   // Commands DOWN
-  CBOROtaUpdateCmdDown    = 0x010100,
-  CBORThingUpdateCmd      = 0x010400,
-  CBORThingDetachCmd      = 0x011000,
-  CBORLastValuesUpdate    = 0x010600,
+  CBOROtaUpdateCmdDown     = 0x010100,
+  CBORThingUpdateCmd       = 0x010400,
+  CBORThingDetachCmd       = 0x011000,
+  CBORLastValuesUpdate     = 0x010600,
   CBORTimezoneCommandDown = 0x010900,
```
andreagilardoni, line 38:
```suggestion
  CBORTimezoneCommandDown  = 0x010900,
```

fabik111: Done

andreagilardoni:
Hi @fabik111, I would like to avoid adding another entangled dependency to `ConnectionHandler` in the Device sub module. One possible way to avoid this is to add a `std::function<void(NetworkSetting*)>` as a `IoTCloudDevice` attribute with a corresponding `setNetworkSettingCbk(...)` method. In the future this could be changed and used to pass generic Device settings to the cloud, ie. fw version etc.

andreagilardoni (approved):

PR #564, opened 2025-07-11
Force sending the network configuration when it's updated via Serial

fabik111:
The current version does not transmit to the cloud the new network configuration after it has been update with the NetworkConfigurator via Serial.

The PR adds a new state `UpdatePhy` for updating the network configuration. When the network configurations are changed the board is put in `Disconnect` state.


andreagilardoni (approved):

pennam (approved):

pennam (approved):
:100:

PR #581, opened 2025-10-23
Prov sketch bin 0.6.x

fabik111:


fabik111:
Bin v 0.6.2 used for Portenta H7, Nicla Vision, Giga

PR #594, opened 2026-03-25
Fix Provisioning 2.0 sketch for production of Opta Lite and Opta rs485

fabik111:
Since the Opta Lite and Opta RS485 are not equipped with the WiFi module, the factory doesn't upload the WiFi FW file, but only adds the partition to store it. So, the board at the production test stage triggered the partition of the QSPI flash. This was not expected.

Changes:
- update the [Arduino_CloudUtils](https://github.com/arduino-libraries/Arduino_CloudUtils/pull/36) dep to the version which fixes the check of the QSPI Flash handling the missing WiFi FW file for Opta Lite and Opta RS485 
- send the WiFi FW version only for Opta WiFi


andreagilardoni:
Moved the check for wifi module presence in Connection Handler library: https://github.com/arduino-libraries/Arduino_ConnectionHandler/pull/165

andreagilardoni:
This PR requires https://github.com/arduino/ArduinoCore-mbed/pull/1110 to be released in order to properly work

.github/workflows/compile-provisioning.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         - name: Arduino_SecureElement
           version: 0.4.0
         - name: Arduino_CloudUtils
-          version: 1.1.2
+          version: 1.1.3
```
andreagilardoni: The latest version available is 1.2.0 which adds zephyr support only, so shouldn't affect this

.github/workflows/compile-provisioning.yml
```diff
@@ -121,6 +121,8 @@ jobs:
               - name: arduino:mbed_nicla
                 version: 4.3.1
             libraries: |
+              - name: ArduinoBearSSL
+                version: 1.7.6
```
andreagilardoni: the latest version available is 1.8.0

andreagilardoni (changes requested):

andreagilardoni (approved):

